PR #13, opened 2020-05-08
Add the IntelliJ files to .gitignore

bschmalhofer:


svenoe:
I will discuss with Stefan how to handle files from external programs. In principle anyone could use .git/info/exclude to ignore files from their individual programming environment etc.

svenoe:
Could be added at a later stage.

PR #14, opened 2020-05-08
Issue 12 cpanfile take 2

bschmalhofer:
A litte bit of fiddling with otobo.CheckModules. Allow the option --cpanfile when an user wants it.


bschmalhofer:
It's probably smarter to remove the check whether a module is already available.

PR #22, opened 2020-05-13
Isssue 16 psgi 2

bschmalhofer:
Moved bin/cgi-bin/app.psgi to bin/psgi-bin/otobo.psgi


bschmalhofer:
The goal is to adapt only otobo.psgi, no changes in the other modules.

PR #83, opened 2020-06-08
Issue 82 as password

bschmalhofer:


bschmalhofer:
Hi, I have tested this with MariaDB. Could you guys take a look at MySQL 5.7 and MySQL 8, in order to be on the safe side.

bschmalhofer:
Fixed the check for MariaDB and swapped the two cases.

Issue #122, opened 2020-07-01
Docker: Support for Redis


bschmalhofer:
Add a container running Redis and adapt the config so that the Redis Cache is used.
It propably makes sense to put the Redis workfiles on the host.

bschmalhofer:
Seems to work.

PR #124, opened 2020-07-01
Issue #122: first stab as adding Redis support

bschmalhofer:


bschmalhofer:
Looks good. Calling SCAN with the redis-cli in the container redis gave me some OTOBO related keys.

PR #130, opened 2020-07-02
Issue-78: fix unicode issues

bschmalhofer:
Mysterious interaction between 'unicode_eval', 'use utf8' and 'binmode'.
Two takeaway messages:
 - The UTF8 flags matters when doing string comparisons
 - Evaling files that were read in with binmode end up
   with strings that have the expected bytes but the UTF8 flag turned off


bschmalhofer:
Still many failures, but mostly not unicode related.

Issue #163, opened 2020-07-08
UnitTests: failure in ./scripts/test/Package/ConfiguredRepositoryDefinitionGet.t


bschmalhofer:
Looks like no repository is configured. Investigate.

bschmalhofer:
The file  _Kernel/cpan-lib/Mozilla/CA/cacert.pem _ was not copied into the container.

bschmalhofer:
Look good. Closing.

PR #171, opened 2020-07-09
Issue #163: make exceptions for actually needed crypto files

bschmalhofer:
Being overly cautious has it's price.


bschmalhofer:
Test is now succeeding.

Issue #213, opened 2020-07-17
Kernel/System/SysConfig/Migration.pm does not compile


bschmalhofer:
Fix this module if it is still needed.

`# Can't locate scripts/MigrateFromOTRS/Base.pm in @INC (you may need to install the scripts::MigrateFromOTRS::Base module) (@INC contains: blib/lib /home/bernhard/devel/OTOBO/otobo /home/bernhard/devel/OTOBO/otobo/Kernel/cpan-lib /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.30.0 /usr/local/share/perl/5.30.0/x86_64-linux-gnu-thread-multi /usr/local/share/perl/5.30.0 /usr/lib/x86_64-linux-gnu/perl5/5.30 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.30 /usr/share/perl/5.30 /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base /home/bernhard/devel/OTOBO/otobo /home/bernhard/devel/OTOBO/otobo/Kernel/cpan-lib /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.30.0 /usr/local/share/perl/5.30.0 /usr/lib/x86_64-linux-gnu/perl5/5.30 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.30 /usr/share/perl/5.30 /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base) at /home/bernhard/devel/OTOBO/otobo/Kernel/cpan-lib/parent.pm line 16.
`

bschmalhofer:
This is related to #190.
Sven, could you take a look at it? As I never looked into the migration procedures. Is migration from OTRS 5 actually supported? Also there seems to be some mixup 'OTRS' vs. 'OTOBO'. There is no such thing as OTOBO 5 or OTOBO 6.

bschmalhofer:
Set the milestone to 10.1 because this should be an easy fix.

bschmalhofer:
Base.pm is in _Kernel/System/MigrateFromOTRS_ not in _scripts/MigrateFromOTRS_. The package name was adapted. Closing.

StefanRother-OTOBO:
Hi Bernhard,

I think you can remove this UnitTest, cause I think it's used for upgrading from 5 to 6. 

Best wishes,

Stefan

[...]

PR #214, opened 2020-07-17
Issue 174 test compile

bschmalhofer:
Let's keep #213 open until we are confident that the changes in Defaults.pm has no bad consequences.


bschmalhofer:
It was decided to roll back the changes for setting up @INC in Defaults.pm. It looks too risky.

PR #274, opened 2020-07-30
Issue 181 clean up usage of env

bschmalhofer:
This is a largish and a bit risky  PR. I'm pushing it now so that problems will surface while developing the other features for OTOBO 10.1.1.


bschmalhofer:
Decline, should be merged with rel-10_1

PR #341, opened 2020-08-10
Issue 136 handling of xml comments

bschmalhofer:
This PR should be reviewed before it is applied. Special attention please for XML.t, as the behavior was changed in some edge case. Of course IMHO the behavior is improved in the patch.


bschmalhofer:
Thought of an simplification.

stefanhaerter:

Issue #534, opened 2020-10-12
MigrateFromOTRS: streamlined migration where otrs db is on same MySQL/MariaDB db server


bschmalhofer:
A common use case is when the otrs-database is on the same server as the otobo-database. This case should be optimised.

TODO:

- [x] merge the adaptions for migration from rel-10_1 to rel-10_1
- [x] document the proposed general approach
- [x] document copying of databases with mysqldump
- [x] document how to do the streamlined migration under Docker
- [x] implement copying of tables with batch insert
- [x] support column shortening for the batch insert
- [x] implement moving of tables with ALTER TABLE ... RENAME ...
- [x] dump 'SHOW CREATE TABLE' in case of problems
- [x] adapt the charset of the copied table, including shorting of char fields to 190 characters where needed, see #542 
- [x] checkbox for the case when tables can be moved instead of copied
- [x] limit the RENAME approach to Docker-based installations
- [x] prepare SQL dumps for OTRS schema and data in _scripts/backup.pl_

bschmalhofer:
One problem to consider: the table name 'groups' is not allowed in MySQL 8. But we got lucky here. According to https://mariadb.com/kb/en/reserved-words/ 'groups' is not reserved on MariaDB 10.

bschmalhofer:
A test migration with the batch insert completed now. But  there are still thinks to do.

bschmalhofer:
A streamlined migration worked at least once, on a simple test database.

bschmalhofer:
The shortening of columns currently only works for the non-batch copying. Therefore shortened colums trigger a fallback to the non-batch approach. Make shortening work for all cases.

The fallback because of 'DirectBlob' is not relevant here. It's only an issue when the database system is changed, e.g. MySQL to Postgres,

bschmalhofer:
This is very strange. Why is the connection lost when doing a simple CONVERT TO CHARACTER SET ?
Note that the column *customer_user_login* is already shortened to 191 chars and that foreign key constraints are already removed.


```
MariaDB [homebernharddeve]> show create table service_customer_user;
+-----------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table                 | Create Table                                                                                                                                                                                                                                                                                                                                                              |
[...]

PR #537, opened 2020-10-12
Issue #534 simple mysql migration

bschmalhofer:
Pull in changes from the rel-10_1 branch.


bschmalhofer:
The commits done on 2021-06-09 were actually intended for #573.

Issue #444, opened 2020-09-15
PSGI: test /hello and /otobo/dbviewer


bschmalhofer:
/dbviewer tends to bitrot. Add a test that runs only  in the Docker use case.

bschmalhofer:
It might be less confusing when two test scripts are created: Hello.t and DbViewer.t.

bschmalhofer:
Improvements of the test suite can be done in the rel-10_0 branch. Do that and move the milestone to 10.0.6

bschmalhofer:
The tests was not working because `Kernel::System::UnitTest::Helper` has RestoreDatabase enabled. This means that the test user was created in a transaction which is never committed. Thus the webserver can't see the added user.

bschmalhofer:
Tests are looking good. Closing this issue.

PR #571, opened 2020-10-23
Issue #444 dbviewer hello in 10 0

bschmalhofer:


bschmalhofer:
Merge the current state, even though DbViewer.t still has TODO tests.

Issue #500, opened 2020-09-30
Review code of the migration modules


bschmalhofer:
@StefanRother-OTOBO asked me to take a look at the code in _Kernel/System/MigrateFromOTRS_ .
Encountered problems or missing features should get their own issues.

bschmalhofer:
The subs _CheckOTOBOVersion() and _CheckOTRSVersion() are each declared in two different modules. Is the check performed twice?

bschmalhofer:
Closing this issue. However further improvements to the migration modules are appreciated.

PR #579, opened 2020-10-26
Issue #500 review migration modules

bschmalhofer:
@svenoe  Hast du mal einen frischen Blick darauf? Vor allen ob Variablen mit den Pfaden von Config.pm und RELEASE vernünfig benamst sind.


svenoe:
I looked at it for a bit, no testing done. Looks fine to me. One could think about renaming the subroutines to what they are intended to do on a program-logic level, not to the "how they do it", but I guess that was not clearer before, so no objections here, and I'm also not 100% sure, what that even is. (I guess it's just a check, if basic access to the OTRS and OTOBO files works.)

PR #611, opened 2020-11-02
CustomerDashboard TicketState

ghost:
Loader::Module::CustomerDashboard###002-CustomerDashboard -> CSS
Frontend::Customer::View::Dashboard::Tiles -> 
otobo


svenoe:
Hey Dennis,

thank you for your contribution! :)

I had a quick look on the files. I will need some time to include them into otobo - I have to discuss on how exactly we want to do stuff like this with the others, and I probably want to change it a bit, use the files we already have, probably add a config option to only select the active tickets of the last x days or so (else in the end everyone will have a big blue circle), maybe add something like a maximum, and then test it a bit.

Sven

svenoe:
Thanks for updating. I hope I will find some time soon... :)

Issue #627, opened 2020-11-09
MigrateFromOTRS: specific message when DB connect is possible, but there are no tables


bschmalhofer:
For testing I created a Postgres user and an empty schema. But I became confused because the message said that the connect was not possible. The real reason why the check failed was that there were no tables.

bschmalhofer:
Enhanced the error message when the sanity check fails.

bschmalhofer:
The messages emitted when the hostname is incorrect, or when there are no tables, is now more sensibel. Closing the issue.

PR #629, opened 2020-11-09
Issue #627: enhance the reporting from SanityCheck

bschmalhofer:
add explicit check whether the database connect works


bschmalhofer:
Print a more specific message in the sanity check.
Explicitly check for the database connect.

PR #634, opened 2020-11-10
Fix email validation when email is kinda ok

niccord:
Hi everyone!
The method `Email::Valid->address( $Param{Address} )` returns `undef` if the parameter is not a valid email, otherwise it returns the email "validated" so without spaces, for example.

If you set "test@gmail .com" as a recipient, the method above returns a string: the email "validated", so "test@gmail.com".
Doing so, the `if` check in the Perl code considers it valid and goes on.
The send fails because of the space in the recipient email.

With the proposed change, it will no longer fail when sending "kind of correct" email addresses as the one above.

### DIFF
--- a/Kernel/System/CheckItem.pm
+++ b/Kernel/System/CheckItem.pm
@@ -122,8 +122,9 @@ sub CheckEmail {
     # remove comment from address when checking.
     $Param{Address} =~ s{ \s* \( [^()]* \) \s* $ }{}smxg;
 
+    $Param{Address} = Email::Valid->address( $Param{Address} );
     # email address syntax check
-    if ( !Email::Valid->address( $Param{Address} ) ) {
+    if ( !$Param{Address} ) {
         $Error = "Invalid syntax";
         $Self->{ErrorType} = 'InvalidSyntax';
     }


svenoe:
Hi Nicola, thanks again. :)

This time I'm not sure whether I understand what you want to achieve with this. I understand that you could in principle use the return value of Email::Valid->address() as the "fixed" email address, which in principle is not a bad idea in my opinion, too, but you are replacing this just in the OTOBO-method CheckEmail(), which doesn't return the email address further. This means at no place you will really replace the email, but just return a true (1), where it is not (but the email address could be fixed). One could think about changing the behavior of CheckEmail() itself, but that would be a major change, which would have to be discussed in beforehand. If I'm missing something, please explain it to me!

niccord:
Oh shoot, I miss that part.
You're absolutely right: I think the behaviour of CheckEmail should be discussed, anyway I can provide a workaround for this bug (later).

svenoe:
Ok, then I will close this pr.

I suggest, that before starting to put time into coding here, you should open an issue stating what exactly you want to achieve, and how you plan to do this, maybe too, if a change of the functionality of CheckEmail() will be involved. The thing is, that this is not a minor change - if you do a "grep -r 'CheckEmail(' Kernel", there are quite a lot of places which are affected, and that doesn't include potential usage in various packages...

Thanks anyways, we are happy to see interest and receive contributions from various places! :)

Issue #644, opened 2020-11-21
MigrateFromOTRS: migrate database using the commands mysqldump and mysql


bschmalhofer:
In #534 an migration strategy was implemented that relied on moving table between database schemas using "ALTER TABLiE ... RENAME TO ...". This seemed to work in the considered case. However the considered cased is the well behaved cased where there are just about no differences between the source and the target schema. In real migrations there are problems when some of the attributes types have diverged. This divergence causes that some foreign key relationships can't be recreated during the migration process.

See also #575.

An example:

In a regular OTRS 6 or OTOBO 10 Installation:
```
CREATE TABLE auto_response ( 
  id smallint(6) NOT NULL AUTO_INCREMENT,
```
At the customers site:
```
CREATE TABLE auto_response ( 
  id INTEGER NOT NULL AUTO_INCREMENT, 
```
Leads to the error:
`errno: 150 "Foreign key constraint is incorrectly formed"`

Probably depending on the order in which the tables are migrated these error can occur when renaming the table or when re-adding foreign key constraints.

The suspicion is that these kind of errors might also occur with other migration strategies.

This problem was discussed by @StefanRother-OTOBO  and @bschmalhofer . The conclusion was to abolish the 'RENAME TO' solution and to provide support for a database  migration strategy based on mysqldump, sed, and mysql.

TODO:

- [x] Make the OTOBODatabaseMigrate step in migration.pl optional
- [ ] Provide instructions and enhance _scripts/backup.pl_ for a direct database migration
- [x] consider charset changed to utf8mb4
- [x] avoid double replacements of the the charset
- [ ] rename the database to the DatabaseName from _Kernel/Config.pm_
- [x] consider skipped tables, maybe checking foreign key relationship of these tables
- [x] consider renamed tables
- [x] consider  InnodB max key size in MySQL 5.6 or earlier: 191 = int( 767 / 4 )
- [x] don't consider DirectBlob conversion
- [x] don't check for concurrent migrations
- [x] don't limit to Docker based installations
- [x] no thorough check of the migrated schema
- [x] no support for restarting interrupted migrations
- [x] do not reset auto increment fields, as this is not neccessary for MySQL
- [x] allow to skip Database copy in migration.pl

bschmalhofer:
@StefanRother-OTOBO 
I have addrf the generation of the SQL scripts to _scripts/backup.pl_ . The changes is available in the pull request #645.
[...]

PR #645, opened 2020-11-21
Issue #644 migrate with mysqldump

bschmalhofer:


bschmalhofer:
This PR won't be merged as there was a major mess when rel-10_1 was merged into the feature branch.
See #646 instead.

Issue #175, opened 2020-07-09
UnitTests: Reactivate Selenium tests


bschmalhofer:
Looks like testing with Selenium is deactivated.

wollmers:
If Mojolicious is installed, `Mojo::UserAgent` can be used for many tests under perlish test infrastructure.

bschmalhofer:
Yes, we could also create a Mojolicious wrapper for otobo.psgi and use https://metacpan.org/pod/Test::Mojo.

svenoe:
There are quite a number of Selenium tests present however, which should be used again, at some point!
(If there are reasons to swap to Mojolicious for newer tests I'm not against it. It would just be a waste if we didn't use what's there already, I figure...)

bschmalhofer:
Selenium testing is activated by pasting Selenium configuration into Config.pm

    # ---------------------------------------------------- #
    # setting for running Selenium tests under Docker      #
    # ---------------------------------------------------- #
    $Self->{'SeleniumTestsConfig'} = {
        remote_server_addr  => 'localhost',
        port                => '5555',
        platform            => 'ANY',
        browser_name        => 'firefox',
        extra_capabilities => {
            marionette     => \0,   # Required to run FF 47 or older on Selenium 3+.
        },
    };

Current the setup is such that a Selenium or PhantomJS server is required. On first sight it looks like "Without Standalone Server" variant would be more convenient. Trying that variant looks like a worthwhile experiment. The prerequisite is that `geckodriver` is installed. Gecko provodis the protocol Marionette for remote controlling the Browser.

wollmers:
@bschmalhofer Where are the Selenium tests running in the above case? In the docker host? I guess, because it needs the binary of FireFox.

Tried yesterday to get a Selenium client running via Perl on Mac OS. Specifying `firefox` doesn't work on Mac OS even with full path (wich starts FF from the command line). The modules for Selenium on Cpan do not look well maintained. 

Using PhantomJS (without FireFox) is very slow for me even with a local website reacting in under 0.1 secs.  Seems to start PhantomJS for each request or has a builtin delay.

```
#!perl

use strict;
use warnings;

use Selenium::PhantomJS;

my $driver = Selenium::PhantomJS->new(
	'binary' => '/usr/local/bin/phantomjs',
);

$driver->get('http://127.0.0.1:3000/line/ONB_aze_19110701_001.jpg_tl_43.gt.txt');
print $driver->get_title(),"\n";
$driver->quit();
[...]

PR #661, opened 2020-12-05
Issue #175: add the standard pragmas

bschmalhofer:
Also enhance the POD


bschmalhofer:
The syntax check action works again.

Issue #652, opened 2020-11-30
QuickClose error when Article Option is active


Wolverine80:
Error
` Can't locate object method "ArticleCreate" via package "Kernel::System::Ticket::Article" at /opt/otobo/Kernel/Modules/AgentTicketZnuny4OTOBOQuickClose.pm line 89.\n`
when trying die QuickClose a ticket when the Option to create an article text is active. 

AgentTicketZnuny4OTOBOQuickClose.pm seem so call the wrong CreateArticle function. 

I changed a line in attached patch to make it work. I don't know if it's the right function though. #notaprogrammer

Wolverine80:
Forgot the patch.

[AgentTicketZnuny4OTOBOQuickClose.pm AgentTicketZnuny4OTOBOQuickClose.pm.txt](https://github.com/RotherOSS/otobo/files/5615198/AgentTicketZnuny4OTOBOQuickClose.pm.AgentTicketZnuny4OTOBOQuickClose.pm.txt)

svenoe:
Thanks Wolverine, I'll have a look at it!

svenoe:
Implemented - thanks again.

PR #664, opened 2020-12-09
use the correct article backend

wollmers:


wollmers:
See issue https://github.com/RotherOSS/otobo/issues/652

Issue #691, opened 2020-12-29
Testing: Selenium/Output/PDFTicket.t: Use of uninitialized value


wollmers:
Without Selenium configured:

```
[20:46:18] /opt/otobo/scripts/test/Selenium/Output/PDFTicket.t......................................................... 
[Tue Dec 29 20:46:19 2020] PDFTicket.t: Use of uninitialized value in string ne at /opt/otobo/scripts/test/Selenium/Output/PDFTicket.t line 28.

# Seeded srand with seed '20201229' from local date.
ok 1 - PDF test currently only supports Firefox
1..1
ok      888 ms ( 0.00 usr  0.01 sys +  0.80 cusr  0.09 csys =  0.90 CPU)
```

bschmalhofer:
@wollmers I wondered why the issue is not referenced in the commit message. Looks like the prepare-commit handler is not installed. See `bin/otobo.Console.pl Dev::Git::InstallHooks --help`

wollmers:
@bschmalhofer Good to know that something like `bin/otobo.Console.pl Dev::Git::InstallHooks` exist, as I wouldn't expect or search it there, nor is it convenient for my preferred workflow. I will just copy the `prepare-commit-msg` hook into `.git/hooks`.

PR #692, opened 2020-12-29
Testing: avoid uninitialized value #691

wollmers:


wollmers:
Solves https://github.com/RotherOSS/otobo/issues/691

Issue #632, opened 2020-11-10
Check whether the input "Expand" for the different forms is still needed, remove it if not.


svenoe:
In CustomerTicketMessage.tt and others there is a hidden input named "Expand". This was introduced with Type, Service and SLA, and if I understand it correctly it is the ancient way of an ajax update. I am not sure though, why this is still there, and whether this is still used somewhere. (Also I'm a bit confused, why it is handled as "Error" in the respective perl modules, e.g. CustomerTicketMessage.pm, but I guess this was just convenient back then...)

For reference see commit 6f7e6a6b4bfff97279538d29a971f8b22b4c8b21 of otrs for its first introduction, and eade32018474721e02db944319c582f450674162 which is a commit which changes where the expand input is set. I have not further looked into it, yet.

bschmalhofer:
I looked at the code and my impression is also that Expand can be removed. Let's check whether there are similar cases.

bschmalhofer:
More greping on OTRS and OTOBO source code. Here are the cases that need to be investigated:

- [x] ExpandCustomerName[12]? in AgentTicketEmail and AgentTicketPhone:

```
./Kernel/Modules/AgentTicketEmail.pm:        my $ExpandCustomerName = $ParamObject->GetParam( Param => 'ExpandCustomerName' )
./Kernel/Modules/AgentTicketEmail.pm:            my $Item = $ParamObject->GetParam( Param => "ExpandCustomerName$Number" ) || 0;
./Kernel/Modules/AgentTicketPhone.pm:            my $Item = $ParamObject->GetParam( Param => "ExpandCustomerName$Count" ) || 0;
./Kernel/Modules/AgentTicketPhone.pm:        my $ExpandCustomerName = $ParamObject->GetParam( Param => 'ExpandCustomerName' )
./Kernel/Output/HTML/Templates/Standard/AgentTicketPhone.tt:                <input type="hidden" name="ExpandCustomerName" id="ExpandCustomerName" value="0"/>
./Kernel/Output/HTML/Templates/Standard/AgentTicketPhone.tt:#    <input type="hidden" name="ExpandCustomerName" value="1"/>
./Kernel/Output/HTML/Templates/Standard/AgentTicketEmail.tt:                <input type="hidden" name="ExpandCustomerName" id="ExpandCustomerName" value="0"/>
[...]

PR #753, opened 2021-01-25
Issue #632 historic autocomplete

bschmalhofer:
This is just a preview of the adaptions for #632. This PR should not be applied yet.


bschmalhofer:
This PR won't be merged. Let's start over and consider only Expand.

PR #758, opened 2021-01-27
Fix HTTP::REST Transport error if request body is empty

sternenseemann:
In OTRS 6 if you'd perform an request using the HTTP::REST backend that had a empty request body (i. e. `{ Data => {}, }`), it would fail with this error message:

```
Not a SCALAR reference at /opt/otrs/bin/cgi-bin/../../Kernel/cpan-lib/LWP/Protocol/http.pm line 256.
```

This is caused by `REST::Client` missing its body argument. The solution is to pass `undef` as body argument if our body turns out to be empty.

This issue is still present in OTOBO and can be fixed by the very same patch which has been taken from <https://github.com/OTRS/otrs/pull/2028> where you may find additional context.

### DIFF
--- a/Kernel/GenericInterface/Transport/HTTP/REST.pm
+++ b/Kernel/GenericInterface/Transport/HTTP/REST.pm
@@ -760,8 +760,16 @@ sub RequesterPerformRequest {
     }
     push @RequestParam, $Controller;
 
-    if ( IsStringWithData( $Param{Data} ) ) {
-        $Body = $Param{Data};
+    # Only POST, PUT or PATCH have a body. If it is empty
+    # (i. e. $Param{Data} = {}), undef is passed to REST::Client.
+    if ( $RestCommand eq 'POST'
+         || $RestCommand eq 'PUT'
+         || $RestCommand eq 'PATCH'
+         ) {
+        if ( IsStringWithData( $Param{Data} ) ) {
+            $Body = $Param{Data};
+        }
+
         push @RequestParam, $Body;
     }
 


svenoe:
Hi sternenseemann,

sorry for the late reply - we will implement this, thank you for your contribution! I'm not sure yet, where, though. We, too, have made a couple of minor tweaks to the generic interface in some packages, which have to be included into the standard. I'm just hesitating a bit, because at one point, we really need to introduce better error handling, too, and I was thinking about starting a larger overhaul.

Anyways, we have to decide on that - thanks again, Sven :)

sternenseemann:
> I was thinking about starting a larger overhaul

Are you also interested in introducing new/missing features to the `GenericInterface` in the mid/long term?

There are some larger oversights/missing features in `GenericInterface` at the moment I'd say, for example there's no way to set headers for `Requester`s at the moment — a feature I had to patch in for an extension of mine. I generally am interested in implementing this properly (with support in the WebService admin interface etc.), but held off because I felt like this was never gonna be merged into OTRS 6.

svenoe:
> Are you also interested in introducing new/missing features to the `GenericInterface` in the mid/long term?

Definitely. And we are happy about any contribution! :)

I opened https://github.com/RotherOSS/otobo/issues/772 - please feel free to write suggestions or provide further pull requests. (We cannot promise to handle everything within days, but eventually things will find their way.)

svenoe:
A smaller change then I had remembered it to be, and quite independent of everything else I guess. :)

Thanks again and sorry for the delay.

svenoe:
Ah...messy... I accidentally merged into the #772-branch. I cherry-picked this commit into rel-10_0 now, too, though. (34c0a55a8e489598c5a2d7f76fd119fd4047bb91)

sternenseemann:
Sorry, changed the target branch after reading <https://github.com/RotherOSS/otobo/issues/956#issuecomment-825463874> without giving you a heads up.

PR #811, opened 2021-02-19
fix for GenericInterface: If one ticket has less articles than a give…

reneeb:
…n article limit, the article limit is set to the number of articles in that article.

And that new article limit is taken for all subsequent tickets as well. See https://github.com/znuny/Znuny/issues/16


svenoe:
Thanks Renee. :)

reneeb:
https://github.com/reneeb/Znuny/blob/reneeb_16/scripts/test/GenericInterface/Operation/Ticket/TicketGetBulkWithArticleLimit.t is a unittest script for the issue. Feel free to copy it... You just merged the PR a few seconds before I was able to add it to the PR ;-)

PR #829, opened 2021-03-01
Fix missing colon on css top

niccord:
Hi there,
I've found a small bug on the CSS of Core.Login.css and I've fixed it right away.

Cheers.


svenoe:
Thanks Nicola! :)

PR #844, opened 2021-03-05
Error description in support package generation

### LINKED ISSUE #2204: enhance description in the error popup in the Support Data Collector
This is from an old pull request by [massicomp](https://github.com/massicomp):

> i've added some description in the error popup in the Support Data Collector section when it fails to produce a Support Package.
> Actually, the popup shows the message "It was not possible to generate the Support Bundle.", with this push it will show the same string and the message returned from the GenerateSupportBundle subaction.
> 
> I think it might be useful for the system admin to understand the problem with the installation.

I have a added that to OTOBO 11 as it is not an urgent fix, but an useful improvement.

massicomp:
Hi :)

i've added some description in the error popup in the Support Data Collector section when it fails to produce a Support Package.
Actually, the popup shows the message "_It was not possible to generate the Support Bundle._", with this push it will show the same string and the message returned from the GenerateSupportBundle subaction.

I think it might be useful for the system admin to understand the problem with the installation.

Regards

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.Admin.SupportDataCollector.js
+++ b/var/httpd/htdocs/js/Core.Agent.Admin.SupportDataCollector.js
@@ -102,7 +102,7 @@ Core.Agent.Admin = Core.Agent.Admin || {};
                     TextClass = 'Error';
 
                     Core.UI.Dialog.ShowContentDialog(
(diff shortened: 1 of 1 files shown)


svenoe:
Hi Massimo,

quite a late reply and rather tidying. I'll close this pr, in principle it might be useful, but it lacks html escaping. See also my comment in the issue.

stefanhaerter:

Issue #814, opened 2021-02-19
Allow article cache in Redis


niccord:
Currently, you can only have articles cache on Memcached:
```perl
return $Self if $CacheModule ne 'Kernel::System::Cache::MemcachedFast';
```
by removing this row, you can achieve article storage on filesystem and _caching onto Redis_, which becomes incredibly faster than without caching.

https://github.com/RotherOSS/otobo/blob/cacfb8a8547ed79e16aa322e2b873dc36d8bce91/Kernel/System/Ticket/Article/Backend/MIMEBase/ArticleStorageFS.pm#L83

svenoe:
Hi niccord,

thanks for the suggestion. Simply removing this line is not a valid solution as there are still systems running without Redis. Also it should be tested first, how big the impact of this on memory usage is. But of course we can enable caching specifically for Redis, too, here, and it is definitely something we should look into. :)

Best regards, Sven

svenoe:
I merged the pr, thank you.

I will also close this issue, I would be interested in some benchmarks of how much faster various actions get, with this. If anyone has data, by chance... ;)

PR #849, opened 2021-03-06
Allow article cache on Redis

niccord:
As proposed, this will allow users to cache articles on Redis (#814)


bschmalhofer (commented):
I didn't dig deep into this issue, but should'nt article caching be allowed on both Memcached and on Redis? Or course depending in which caching module has been set up. So the the code should be changed to:

```
    my $CacheModule = $ConfigObject->Get('Cache::Module') || '';
    my %CacheModuleIsSupported = (
        'Kernel::System::Cache::MemcachedFast' => 1,
        'Kernel::System::Cache::Redis' => 1,
    );
    return $Self unless $CacheModuleIsSupported{ $CacheModule };

```

niccord:
Your idea is better than mine, of course 😁 
Cheers

svenoe:
I was a bit hesitant at first, because I didn't want to change this without a config option at least, but really I just didn't read properly, this config option has to be set via Kernel/Config.pm manually anyways, I guess. (Cache::ArticleStorageCache two lines above...^^)
So I see nothing speaking against this.

PR #865, opened 2021-03-15
Added the possibility to use Core.AJAX.FormUpdate for multiple backend-actions on one page

frankhildebrandt:
Currently we are working on a Widget for the AgentTicketZoom-Page which uses CustomFields and a custom Custom Endpoint. Unfortunate the current Frontend uses by definition the Action used in the BrowserUrl. My Change trying to resolve this issue, by searching for an Input-Field named "Action" in the Scope of the EventElement and using it's value.

### DIFF
--- a/var/httpd/htdocs/js/Core.AJAX.js
+++ b/var/httpd/htdocs/js/Core.AJAX.js
@@ -569,6 +569,9 @@ Core.AJAX = (function (TargetNS) {
             Data = GetAdditionalDefaultData(),
             QueryString;
 
+        $EventElement.find('input[name="AJAXAction"]').each(function () {
+            Data.Action = $(this).val();
+        });
         Data.Subaction = Subaction;
         Data.ElementChanged = ChangedElement;
         QueryString = TargetNS.SerializeForm($EventElement, Data) + SerializeData(Data);


svenoe:
Hi Frank,

in principle I see nothing speaking against providing this functionality. Is there anything speaking against passing `Action` as an argument, from your side, though? I would prefer that for one, because it's more consistent, but more so, because if I didn't overlook something, `Action` is explicitly ignored via `Data` in `TargetNS.SerializeForm()` (line 577 with your change) at the moment, and this might break current functionality for other packages.

Best regards, Sven

(I'm not totally happy, because the list of optional arguments gets quite long, named arguments would definitely be nicer, here. But it is, at it is, I guess... Maybe at some point we should build a wrapper method first and change this in the long run.)

frankhildebrandt:
Reading your comment lets me rethink my subject choice. So let me give you more context on the issue we're trying to deal with.
The widget for the AgentTicketZoom-Page which lets you *display* and *in-place edit* multiple custom-fields embedded in the ticket. Under the current state of `DynamicFields` the `FormUpdate` registration happens within the DynamicField code and AFAIK cannot be altered outside and will default to the current view action (eg. ActionTicketZoom). 

In my case I'm bypassing this problem by embedding the CustomField inside a separate form with an added hidden input field named `Action` to override the action called by the FormUpdate function. I tried to have a solution with as little side-effects as possible. Am I overlooking something?

svenoe:
Ah, you're right, I didn't think about the DynamicFields - still, the hidden Action input is part of 95% of all forms in OTOBO/OTRS, e.g.: https://github.com/RotherOSS/otobo/blob/754877aae66f3c13402d11cde0440f200feddb36/Kernel/Output/HTML/Templates/Standard/AgentTicketPhone.tt#L45

Basically, what your change for probably every form which uses the FormUpdate does, is equal to deleting the following line: https://github.com/RotherOSS/otobo/blob/754877aae66f3c13402d11cde0440f200feddb36/var/httpd/htdocs/js/Core.AJAX.js#L204
As, if it is not defined in `Data`, yet, the Action input will be serialized as part of the form in: https://github.com/RotherOSS/otobo/blob/754877aae66f3c13402d11cde0440f200feddb36/var/httpd/htdocs/js/Core.AJAX.js#L574

---

I was thinking quite a while now on how to solve this, or rather, if it's possible to actually just delete the above line (or probably even the whole `GetAdditionalDefaultData()` method, because at least in the OTOBO standard it is only used from Core.AJAX.js itself, and one could just directly use `GetSessionInformation()` from `FormUpdate()`. This way, what you want would indirectly be implemented without any further code. In the end this boils down to whether you want to be able to do the FormUpdate with a different Action than the submit. But although I'm relatively sure that this is never occurs, I currently don't have the time to go over every occurance of this. I looked into when it was introduced into OTRS, to try to understand the reason for this construct, but it was implemented as it is now as a whole, a bit more than 10 years ago, so that didn't give me any hint either.

Maybe for now, we use your suggestion for 10.0 or 10.1, but change the name of the hidden input field read out to "AJAXAction" or "FormUpdateAction", and open an issue for 11.0 to review this, because I currently assume it's an exception to an unnecessary piece of code, and in the end just clutter...

frankhildebrandt:
A change in name would work perfectly fine for me. Thank you.

Issue #867, opened 2021-03-16
Config Option Frontend::Navigation###ExternalURLJump###1 not possible to change


StefanRother-OTOBO:
Hi,

I can't activate Frontend::Navigation###ExternalURLJump###1, cause missing description.

Thanks,

Stefan

bschmalhofer:
Adding the value in Framework.xml did the trick for a new installation. Installation pre 10.0.10 can upgrade if they want to avoid this effect. Manually editing Framework.xml and running bin/otobo.Console.pl Maint::Config::Rebuild should work too.
Closing this issue.

PR #874, opened 2021-03-18
Issue #867: default value for 'Description' in a SysConfig entry

bschmalhofer:
Specifically for Frontend::Navigation###ExternalURLJump###1.
'Description' is a required parameter for the ExternalJump frontend module.
When it is missing, then this option can't be easily activated in the SysConfig.


bschmalhofer:
Adding the value in Framework.xml did the trick for a new installation. Installation pre 10.0.10 can upgrade if they want to avoid this effect. Manually editing Framework.xml and running `bin/otobo.Console.pl Maint::Config::Rebuild` should work too.
Closing this issue.

PR #885, opened 2021-03-25
Added the possibility to preserve the sorting state on the dashboard

frankhildebrandt:
Currently the dashboard looses its ordering configuration on every reload. As many other views store their order configuration in the user preferences I tried to adapt this behavior.

I'm not sure if this behavior needs to by Enabled/Disabled by sysconfig, or if the user-parameter keys violate some rules.

### DIFF
--- a/Kernel/Modules/AgentDashboardCommon.pm
+++ b/Kernel/Modules/AgentDashboardCommon.pm
@@ -436,6 +436,23 @@ sub Run {
         my $SortBy  = $ParamObject->GetParam( Param => 'SortBy' );
         my $OrderBy = $ParamObject->GetParam( Param => 'OrderBy' );
 
+        my $Direction = $OrderBy;
+
+        if ($ParamObject->GetParam( Param => 'SortingColumn' ) eq $SortBy) {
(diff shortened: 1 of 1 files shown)


svenoe:
Hi Frank,

sorry for the delay, I didn't find much time to look at it the last weeks, and thank you for your contribution.

In general I guess we could add this, although the other views as far as I have seen store only the filters, not the ordering and sorting. The Dashboard differs a little anyways, though. However, we would definitely need a SysConfig-Option for this. I can already see the complaints about overlooked tickets, because people are not aware that they changed their SortBy-option away from age the day before. (This also happens with filters regularly, but there at least you have to do  slightly more than one little click.)

Secondly, there is no explicit naming convention, yet, but at least the camel case with the starting capital letter is used in OTRS/OTOBO everywhere, and looking at the other preferences, your suggestion looks quite different. I would prefer something like "UserDashboardSort-$Name" or similar.

Best regards, Sven

frankhildebrandt:
Sorry for my delay. This week I will update my request to match your requirements.

PR #911, opened 2021-04-06
Configurable index settings for Elasticsearch

io-architect:
Hello!

In CJK area, we sometimes want to change index settings of Elasticsearch(especially analyzer, tokenizer and filter ex: ES's plugin kuromoji(Japan), smartcn(China),  openkoreantext-analyzer(Korea)) because of default analyzer is not work well.
But it is hard-corded and can't change so I make a patch which moved to SysConfig.

There are current settings related index : Elasticsearch::ArticleIndexCreationSettings,  which is a parameter for (only) replication
so I apply template expansion and there are no effects for current users.

Added SysConfig Elasticsearch::IndexSettings is text type JSON-style parameter and can edit if we want.
            {
                "index": {
                    "number_of_shards"   : [% Data.NShards | uri %],
                    "number_of_replicas" : [% Data.NReplicas | uri %]
                },
                "index.mapping.total_fields.limit" : 2000
            }

Regards,
===
Add SysConfig Elasticsearch::IndexSettings.
When createing Elasticsearch index, Use the config JSON.

For backward compatibility of Elasticsearch::ArticleIndexCreationSettings,
Elasticsearch::IndexSettings is treated as template (you can use [% .. %])

### DIFF
--- a/Kernel/Config/Files/XML/Framework.xml
+++ b/Kernel/Config/Files/XML/Framework.xml
@@ -9300,6 +9300,21 @@ via the Preferences button after logging in.
             </Hash>
         </Value>
     </Setting>
+    <Setting Name="Elasticsearch::IndexSettings" Required="1" Valid="1">
+        <Description Translatable="1">Index Settings definition of Elasticsearch.</Description>
+        <Navigation>Core::Elasticsearch::Settings</Navigation>
+        <Value>
+            <Item ValueType="Textarea">
+            {
+                "index": {
+                    "number_of_shards"   : [% Data.NShards | uri %],
+                    "number_of_replicas" : [% Data.NReplicas | uri %]
+                },
+                "index.mapping.total_fields.limit" : 2000
+            }
+            </Item>
+        </Value>
+    </Setting>
     <Setting Name="Frontend::Module###AgentElasticsearchQuickResult" Required="0" Valid="1">
         <Description Translatable="1">Frontend module registration for the agent interface.</Description>
         <Navigation>Frontend::Agent::ModuleRegistration</Navigation>
--- a/Kernel/System/Console/Command/Maint/Elasticsearch/Migration.pm
+++ b/Kernel/System/Console/Command/Maint/Elasticsearch/Migration.pm
@@ -27,12 +27,14 @@ use parent qw(Kernel::System::Console::BaseCommand);
 
 our @ObjectDependencies = (
     'Kernel::Config',
+    'Kernel::Output::HTML::Layout',
     'Kernel::System::CustomerCompany',
     'Kernel::System::CustomerUser',
     'Kernel::System::Elasticsearch',
     'Kernel::System::GeneralCatalog',
     'Kernel::System::GenericInterface::Webservice',
     'Kernel::System::ITSMConfigItem',
+    'Kernel::System::JSON',
     'Kernel::System::Package',
     'Kernel::System::Ticket',
     'Kernel::System::Ticket::Article',
@@ -223,14 +225,14 @@ sub MigrateCompanies {
         );
     }
 
+    my $IndexSettings = $Self->_IndexSettingsGet(%Param);
+    if ( !$IndexSettings ) {
+        # Error is shown in _IndexSettingsGet
+        return 0;
+    }
+
     my %Request = (
-        settings => {
-            index => {
-                number_of_shards   => $Param{NShards},
-                number_of_replicas => $Param{NReplicas},
-            },
-            'index.mapping.total_fields.limit' => 2000,
-        },
+        settings => $IndexSettings,
         mappings => {
             properties => {
                 CustomerID => {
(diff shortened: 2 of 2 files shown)


io-architect:
> 10: 'xmllint' was not found, please install it.

Yes, xmllint is passed in my dev env :)

svenoe:
Hi Tomohisa,

thanks for your contribution! :) We are aware that the Elasticsearch settings (and configurability) are not optimal yet. So far my idea, however, was, to provide a more restricted way of changing stuff, like the analyzer. I think we have to discuss your suggestion of letting the users provide the whole JSON. (It would definitely benefit people who are very familiar with Elasticsearch, but it is much more complicated for people who are not.)

Do you think it would be possible to translate all customizations you want to make to "simple" SysConfig settings? (Like our NShards, or NReplicas I was thinking of a setting Analyzer, etc.) Just out of interest - could you send me an example JSON of how your final IndexSettings should look like, roughly?

(Thinking about it a bit more, I quite like your approach with the templating mechanism. Maybe a combination would be best. We could provide the template in the Default.pm, so it could be changed in the Config.pm (it is a quite basic setting, comparable to the database stuff), and then include more specific options in the SysConfig.)

As I said, I will discuss this, best regards, Sven

(P.S.: The xmllint is a bug in the github action - we are just in the process of setting these up properly.)

io-architect:
Hi Sven, 
Thank you for your quick answer and I'm happy to talk about for OTOBO.

> I think we have to discuss your suggestion of letting the users provide the whole JSON. (It would definitely benefit people who are very familiar with Elasticsearch, but it is much more complicated for people who are not.)

I agree your opinion that to make "simple" SysConfig settings. To put JSON in SysConfig is not smart way as you say.

> Do you think it would be possible to translate all customizations you want to make to "simple" SysConfig settings? 

Sadly I think no. Defining the analysis settings is very deep area I think. Most of OTOBO admins not want complicated config and it is rational but some person may want/need to configure more comprex configs (ex. multilingual search, etc).
To provide the template Default.pm is a good solution I think, too.  

> Just out of interest - could you send me an example JSON of how your final IndexSettings should look like, roughly?

I find some ES's settings template for each languages. 

Europe(+alpha)(Greek, French, Italian, and German, etc..)
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html#english-analyzer
I think the needs to change config is only CJK area's one, but I was surprised at this link :-)

Chinese
https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-smartcn_stop.html
Korean
https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-nori-tokenizer.html
Japanese
https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-kuromoji-analyzer.html

We usually use configs like this. Sorry I don't know Chinese and Korean people write what configs.

`           index => {`
`                number_of_shards   => $Param{NShards},`
`                number_of_replicas => $Param{NReplicas},`
`                analysis => {`
`                    tokenizer => {kuromoji_tokenizer => { type => 'kuromoji_tokenizer', mode => 'search' }},`
`                    filter => { katakana_stemmer => { type => 'kuromoji_stemmer', minimum_length => 4, }},`
`                    analyzer => { default => { tokenizer => 'kuromoji_tokenizer', filter => [ 'katakana_stemmer', 'cjk_width', 'lowercase']}}}}`

There are some patterns but many of them have a bit difference structure X-<

At this time how about to add a simple template in Default.pm?
To reflect admin config values, we need to treverse tree and replace template to values at Maint::Elasticsearch::Migration, 
The code is a bit more complicated but keep config flexible.

Plan A)
`    $Self->{'Elasticsearch::ArticleIndexCreationSettings'} = { # existing`
`        'NS'      => '1',`
`        'NR'      => '0',`
`    };`
`    # Add new config`
`　　$self->{Elasticsearch::IndexSettings} = {`
`　　    index => {`
`   　　　　number_of_shards => '[% Config.NS|uri %]',　# expand when executing`
`　   　　　number_of_replicas => '[% Config.NR|uri %]',`
`        },`
`　　    'index.mapping.total_fields.limit' => 2000,' # or [% Config.FieldsLimit|uri %]`
`};`
Maybe "index.mapping.total_fields.limit" is moved to ArticleIndexCreationSettings like NS/NR (Plan B)

More thinking at the future,  how to treat JSON's array will be a problem. Suppose if we want to add filter or stopwords to SysConfig, a template is like [just_idea.txt](https://github.com/RotherOSS/otobo/files/6291360/just_idea.txt) (future plan C)

Migration.pm will need to be add  something like code:
`$xx->{index}->{filter}->{otobo_keywords} = $ConfigObject->{''Elasticsearch::ArticleIndexCreationSettingsKeyWords'};`
`$xx->{index}->{analysis}->{analyzer}->{default}->{filter} = ...`

This is not cool but well can't be helped. Anyway we still have a flexibility and admin is easy to edit the config.

We should not add Plan C at this moment, because search result may be changed for current users. At leaset these changes are applied not to 10.0.x but 10.1 or future.

Anyway, PlanA (or PlanB) is realistic solution on 10.0.x, I think.
I hope your opinion and I will create and send new pullreq.

svenoe:
Hi Tomohisa,

we discussed this and would be happy about a new pull request. Plan A or B are both fine with me. I agree that in the future this could be extended quite a bit, but it would be too much for a minor release, and in any case, I hope we will be able to build ontop of your pr then, so it's a good solution for now.

Thank you very much for your contribution, Sven

svenoe:
Not needed anymore, as this was commited in https://github.com/RotherOSS/otobo/pull/938 again. Thanks :)

PR #938, opened 2021-04-14
Configurable index settings for Elasticsearch(again)

io-architect:
Hi,

We are discussing Elasticsearch's index definition more configurable in
[PR#911](https://github.com/RotherOSS/otobo/pull/911) and move ESs index setting to Defaults.pm.
For Backward compatibility and extensibility, we  can use '[% %]' style template (like .tt) in new config.

And hard-corded value index.mapping.total_fields.limit is parameterized.

Regards,
Tomohisa


svenoe:
Hi Tomohisa,

thanks again. :) Just as a short info, we are about to release 10.0.10 at the moment (actually our initial plan was last week, so I didn't write here), and we will merge your pr right after that.

Best regards, Sven

svenoe:
Hi Tomohisa,

I merged your pr to a branch first, as we will probably think about some additional adaptions we want to make in the next weeks until 10.0.11 comes out, where it will be included. (See https://github.com/RotherOSS/otobo/issues/961)

Have a nice weekend, Sven

PR #968, opened 2021-04-26
Pass http status from web service to trasport object

### LINKED ISSUE #956: HTTP status code for REST Web Services
Hi folks!
I am trying to implement some REST web service for otobo, but I notice I can only return 200 or 500 as status codes.

After digging into the code a little, I found some case where give different status code is possible, but not handled.

It would be really nice to allow us to specify the HTTP status code, with something like:

``` perl
return {
    Success => 0,
    HTTPStatus => 401,
    ErrorMessage => 'User can\'t access this resource',
};

return {
    Success => 1,
    HTTPStatus => 201,
    Data    => {
        ID => $ID,
    }
};
```

(I already have some working code in my pocket 🤓)

niccord:
Fix #956

### DIFF
--- a/Kernel/GenericInterface/Provider.pm
+++ b/Kernel/GenericInterface/Provider.pm
@@ -38,9 +38,10 @@ use Kernel::System::VariableCheck qw(IsHashRefWithData);
 use Kernel::System::Web::Exception;
 
 our @ObjectDependencies = (
+    'Kernel::GenericInterface::ErrorHandling',
     'Kernel::System::Log',
     'Kernel::System::GenericInterface::Webservice',
-    'Kernel::GenericInterface::ErrorHandling',
+    'Kernel::System::Web::Request',
 );
 
 =head1 NAME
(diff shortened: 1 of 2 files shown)


svenoe:
Yippie - CodePolicy worked! :D 

It still needs to be extended a bit, though... At some point I have to write a small guide - if you download CodePolicy (https://github.com/RotherOSS/CodePolicy) you can run it locally (even without having installed OTOBO) and it will do some further tidying. I commited this in ef243c521246881c4109f9637f26d2a69f0626e1. And also we should enforce a naming convention for the commits (I forgot, myself, because I was thinking about the pr mostly) and have all commits start with "Issue #xxx:" which will cause the commits to be shown in the issues automatically, which is nice, and also enable our little script which writes the CHANGES message to be able to merge those commits.

Anyways, that just as a sidenode - thanks for your contribution, I just merged it. :)

(Edit: Ah, I also changed HTTPStatus to HTTPCode just for consistency.)

PR #970, opened 2021-04-26
Fix CheckEmail for some special cases

### LINKED ISSUE #637: Wrong emails pass validation
Hi everyone,
if you set "test@gmail .com" as a recipient in your ticket reply, the validation passes but the email send returns an error.

This is due to the Perl `Email::Valid->address( $Param{Address} )` function in `Kernel/System/CheckItem.pm` which considers the email valid.

The Perl functions return `undef` if the email is not valid, a string with the "validated" email otherwise, so as the in example above it returns "test@gmail.com".

As a temporary workaround, I propose to check if the output of the function is different from the input address and throw an error in that case.
Something like: 
``` Perl
my $validatedEmail = Email::Valid->address( $Param{Address} );
if ( !$validatedEmail or $validatedEmail ne $Param{Address} ) {
    $Error = "Invalid syntax";
    $Self->{ErrorType} = 'InvalidSyntax';
}
```

You should also consider changing the behaviour of the `CheckEmail` function, returning the validated email address, in a future version.

Edit by @bschmalhofer : Please note the space before the "." in  "test@gmail .com". I starred at the issue for 5 minutes without seeing the space.

niccord:
Fix #637

### DIFF
--- a/Kernel/System/CheckItem.pm
+++ b/Kernel/System/CheckItem.pm
@@ -123,7 +123,8 @@ sub CheckEmail {
     $Param{Address} =~ s{ \s* \( [^()]* \) \s* $ }{}smxg;
 
     # email address syntax check
-    if ( !Email::Valid->address( $Param{Address} ) ) {
+    my $ValidatedEmail = Email::Valid->address( $Param{Address} );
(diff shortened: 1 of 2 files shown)


svenoe:
Quite the delay, but thanks again! :)

PR #977, opened 2021-04-28
Migration redirect stdout

kess-net:
Redirect stdout to stderr for called scripts to prevent their output messing with the regular dataflow. This is relevant for setups using apache but not using mod_perl - for whatever reason.


svenoe:
:)

Issue #945, opened 2021-04-16
Selenium: investigate whether Selenium::Waiter can simplify K::S::U::S::WaitFor()


bschmalhofer:
It could be that the WaitFor() could be simplified. 
Automatic increase of the polling time span is a feature that could be discarded.

bschmalhofer:
This does not work, as the execution of the tests simply halts in the middle of the test suite. Rejecting this request.

PR #992, opened 2021-05-06
Issue #945 selenium waiter

bschmalhofer:


bschmalhofer:
Create this PR just to have a record of the attempted change. This does not fly as the test suits simple halts in the middle of the execution of the tests. It's not worth it to pursue this further. Rejecting this PR.

PR #999, opened 2021-05-10
fixed missing dependencies in workflows

uid66:
Fixed errors in step 'run CodePolicy':
TidyAll::Plugin::OTOBO::Perl::SyntaxCheck
Can't locate Test2/V0.pm in @inc...


svenoe:
Hey uid66,

thanks for the pr! :) The integration of CodePolicy was set on hold for a bit, because there is too much other stuff, atm, but it's definitely needed, so thanks.

Best regards, Sven

Issue #862, opened 2021-03-10
Selenium Testing: no warning when subject is empty


bschmalhofer:
This happens in scripts/test/Selenium/Agent/AgentTicketCompose.t.
The tests expects a warning about an empty subject, but that message is not visible. It is not obvious what the correct behavior ist.
The test dies when the message is not seen. 
The expected message is:

 `       my $Message = 'Article subject will be empty if the subject contains only the ticket hook!';`

Marking the failure as TODO for now.

stefanhaerter:
The warning actually is associated with loading a form draft:

<img width="1256" height="218" alt="Image" src="https://github.com/user-attachments/assets/92a9ecfd-2c42-4781-88d9-a6e29ce5d574" />

I do not really understand what the assertion is trying to check there.

PR #1005, opened 2021-05-12
Issue #862 error disappeared

bschmalhofer:
Actually, the error did not disappear.


bschmalhofer:
Improved the test script, but the failure is still there.

Issue #1014, opened 2021-05-22
Avoid warning about invalid QueueID in TicketQueueOverview.pm line 171.


bschmalhofer:
The warning take from the log of otobo_web_1 is:
`[Sat May 22 20:14:01 2021] plackup: Use of uninitialized value within %Queues in hash element at /opt/otobo/bin/psgi-bin/../../Kernel/Output/HTML/Dashboard/TicketQueueOverview.pm line 171.`

The cause is in line 147:

```
    if ( !@QueueIDs ) {
        @QueueIDs = (999_999);
    }

```
I would have liked to investigate the reasoning for this code, but
https://bugs.otrs.org/show_bug.cgi?id=13142 seems to be down.

bschmalhofer:
lateste test run indicated no problems, PR is merged, closing this issue.

PR #1037, opened 2021-05-27
Issue #1014: avoid the workaround with the dummy queue id

bschmalhofer:
Simply don't search when it is known that there are no results.


bschmalhofer:
The warnings are gone and the test suite runs fine.

@svenoe , could you take a quick look at this?
My understanding is  that the queue ID is guaranteed to not exists, and therefore the results list is guaranteed to be empty.

bschmalhofer:
Discussed this with @svenoe and we concluded that the PR is fine for OTOBO 10.1.x .

PR #1045, opened 2021-05-31
Add LightAdmin function

eyazi:
Adds light admin functionality for these modules: AdminNotificationEvent, AdminTemplate, AdminTemplateAttachment, AdminQueueTemplates, AdminAttachment.

Light admins can only change templates if they have full rw permission on all queues which are affected by the template. If they don't have full rw on all queues, they can only see them without having permission to change/delete them. If they don't have ro on any of the queues, they can't even see the template.

**ToDo: I don't know how to add translations. Someone please add following strings (German translation on the right):**

Template: AdminNotificationEvent
'No permission to edit this ticket notification.' = 'Sie haben keine Berechtigung, diese Ticket-Benachrichtigung zu bearbeiten.'

Template: AdminAttachment
'No permission to edit this attachment.' = 'Sie haben keine Berechtigung, diesen Anhang zu bearbeiten.'

Template: AdminTemplate
'No permission to edit this template.' = 'Sie haben keine Berechtigung, diese Vorlage zu bearbeiten.'

Configuration:
A light admin can be added to one of the above-listed modules by adding the group – in which the light admin is – to the "GroupRo" field in the frontend module registration. The field GroupRo isn't perfect and doesn't fit 100% but it offers out-of-the-box functionality for additional permission controlling. Could be replaced in the future with a self-explaining name.

Example:
- Group/GroupRo: admin-light
Frontend::Module###Admin
Frontend::Navigation###Admin###001-Framework
 
- Group/GroupRo: admin-light
Frontend::NavigationModule###AdminNotificationEvent
Frontend::NavigationModule###AdminTemplate
Frontend::NavigationModule###AdminTemplateAttachment
Frontend::NavigationModule###AdminQueueTemplates
Frontend::NavigationModule###AdminAttachment
 
- GroupRo: admin-light
Frontend::Module###AdminNotificationEvent
Frontend::Module###AdminTemplate
Frontend::Module###AdminTemplateAttachment
Frontend::Module###AdminQueueTemplates
Frontend::Module###AdminAttachment

### DIFF
--- a/Kernel/Modules/AdminAttachment.pm
+++ b/Kernel/Modules/AdminAttachment.pm
@@ -51,6 +51,24 @@ sub Run {
 
         my $Output = $LayoutObject->Header();
         $Output .= $LayoutObject->NavigationBar();
+
+        # Check permission for all linked templates.
+        if ( $Self->{LightAdmin} ) {
+            $Data{Permission} = $StdAttachmentObject->StdAttachmentStandardTemplatePermission(
+                ID      => $ID,
+                UserID  => $Self->{UserID},
+                Default => 'ro',
+            );
+            if ( !$Data{Permission} ) {
+                %Data = ();
+            } elsif ( $Data{Permission} eq 'ro' ) {
+                $Output .= $LayoutObject->Notify(
+                    Priority => 'Notice',
+                    Data => $LayoutObject->{LanguageObject}->Translate( 'No permission to edit this attachment.' ),
+                );
+            }
+        }
+
         $Self->_Edit(
             Action => 'Change',
             %Data,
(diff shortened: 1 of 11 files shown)


svenoe:
Hi Emin,
thanks for the pr. :)
I will most likely merge this into 10.1 instead of 10.0, but I can change that myself.
Sven

StefanAbel-OTOBO:
That already is possible just by adjusting the system configuration. Adding these configurations here is just applying a very specific use case and it also is still missing the proper role.

Edit: sorry, after a talk with @svenoe I found out that this is not about templates in general, but having access to single/specific template in the module. So please ignore my comment.

svenoe:
> That already is possible just by adjusting the system configuration. Adding these configurations here is just applying a very specific use case and it also is still missing the proper role.

No, it's not. You can give access to the admin panels, but you can not give restricted access depending on the group access you have.

stefanhaerter:
I'd like to link this to Issue #2930

svenoe:
We added this to 11.0 in a different pr, because I overlooked this one. Thanks anyways!

PR #1049, opened 2021-06-01
Eliminate otrs reference in tests

### LINKED ISSUE #1013: Eliminate references to OTRS domains in the test suite
The command 
```
$ grep -r "otrs.[com|org|de]" scripts/test/  | grep -v Copyright | grep -v bugs.otrs.org | wc -l
209
```
shows that OTRS domains are still referenced in the test scripts. These could be switched to OTOBO domains. Some test might also make DNS queries to OTRS domains.

niccord:
Fix #1013


svenoe:
See my comment in https://github.com/RotherOSS/otobo/issues/1013

PR #1105, opened 2021-06-30
Show create dates of tickets/articles in the customer portal

eyazi:
As wished from Stefan, this change shows the creation date of tickets and articles in the customer portal if they are older than 24 hours.
Currently, dates are shown as "x days, x hours ago" which can be very confusing.

It might be worth considering adding a setting for this, as not every user might have set up correct time zones for their customers.


stefanhaerter:
Closed via #3614

Issue #1140, opened 2021-07-19
Validation of Dynamic Field from type database not works correctly in the CustomerTicketMessage


OlliW:
when i define a Dynamic Field from type database and it is defined as a mandatory Field in the CustomerTicketMessage.
When i hide the Field with an ACL, the validation on the ACLs not working correctly
I can't create an Ticket from the CustomerTicketMessage, because the mandatory Field is not filled out (it's hidden).

svenoe:
This should be solved with the above commits. We still have some open points for other dynamic field types, and actually the above probably has introduced a regression for those, but this only is that some mandatory fields which are initially hidden and then revealed might skip client side validation, and as all fields are checked on the server side again, and it only affects certain types, this is still an improvement. The rest will be dealt with in #1162.

PR #1152, opened 2021-08-03
Issue #1140: Added Class handling for Database Dynamic Fields and add…

stefanhaerter:
…ed global flag in Regex.

Closing #1140


svenoe:
Check whether fields are really mandatory is missing.

Issue #961, opened 2021-04-23
Enable the system admin to provide complex index settings for Elasticsearch.


svenoe:
@io-architect provided us with a pull request greatly enhancing the ability to customize Elasticsearch. If sensible we should provide one or two more options to adapt stuff in a simpler form via the SysConfig. (https://github.com/RotherOSS/otobo/pull/938)

Please use the branch issue-#961-elasticsearch_index for further development.

kess-net:
Tested and enhanced a little bit.  Branch issue-#961-elasticsearch_index  looks is looking good

kess-net:
merged to rel-10_0

bschmalhofer:
Reopen for fixing _scripts/test/Config/Defaults.t_ .
@kess-net : Could you take a look at the test? It looks like there are divergent settins in Defaults.pm and Framework.xml.

kess-net:
@bschmalhofer , this test is giving me an 'All tests successful.'. What is the problem at your side?

bschmalhofer:
I got a:

> # Failed test 'Elasticsearch::ArticleIndexCreationSettings must be the same in Defaults.pm and and setting default value (is not equal, see below)
> Diff:
> +--+------------------------------+--+------------------+
> |  |Actual data                   |  |Expected data     |
> +--+------------------------------+--+------------------+
> | 1|$VAR1 = \{                    | 1|$VAR1 = \{        |
> * 2|    'FieldsLimit' => '2000',  *  |                  |
> | 3|    'NR' => '0',              | 2|    'NR' => '0',  |
> | 4|    'NS' => '1'               | 3|    'NS' => '1'   |
> | 5|  };                          | 4|  };              |
> +--+------------------------------+--+------------------+
> 
> Actual data:
> $VAR1 = \{
>     'FieldsLimit' => '2000',
>     'NR' => '0',
>     'NS' => '1'
>   };
> 
> Expected data:
> $VAR1 = \{
>     'NR' => '0',
>     'NS' => '1'
>   };
> 
> '
> # at scripts/test/Config/Defaults.t line 194.
> 
> 

after I did a `bin/otobo.Console.pl Maint::Config::Rebuild`. 

Maybe it's an effect of using `bin/docker/quick_setup.pl` instead of using `installer.pl`.

bschmalhofer:
[...]

PR #1165, opened 2021-08-06
Issue #961 elasticsearch index

kess-net:
This should solve https://github.com/RotherOSS/otobo/issues/961. Please review.

### DIFF
--- a/Kernel/Config/Files/XML/Framework.xml
+++ b/Kernel/Config/Files/XML/Framework.xml
@@ -9217,8 +9217,8 @@ via the Preferences button after logging in.
            </Hash>
         </Value>
     </Setting>
-    <Setting Name="Elasticsearch::IndexSettings" Required="0" Valid="0">
-        <Description Translatable="1">Number of shards (NS), replicas (NR) and fields limit for the indices. This replaces 'Elasticsearch::ArticleIndexCreationSettings' in future versions. Use 'Elasticsearch::IndexSettings###...' if you want to define special settings for single indices.</Description>
+    <Setting Name="Elasticsearch::IndexSettings###Default" Required="0" Valid="1">
+        <Description Translatable="1">Number of shards (NS), replicas (NR) and fields limit for the indices. This replaces 'Elasticsearch::ArticleIndexCreationSettings' in future versions. If both are present and not equal this one has priority. Use 'Elasticsearch::IndexSettings###...' if you want to define special settings for single indices.'...' may be one of 'Customer', 'CustomerUser', 'Ticket' or 'ConfigItem.'</Description>
         <Navigation>Core::Elasticsearch::Settings</Navigation>
         <Value>
             <Hash>
--- a/Kernel/Config/Files/XML/Framework.xml
+++ b/Kernel/Config/Files/XML/Framework.xml
@@ -9207,7 +9207,7 @@ via the Preferences button after logging in.
         </Value>
     </Setting>
     <Setting Name="Elasticsearch::ArticleIndexCreationSettings" Required="0" Valid="1">
-        <Description Translatable="1">Number of shards (NS), replicas (NR) and fields limit for the index. Note: 'Elasticsearch::ArticleIndexCreationSettings' is deprecated. For upwards compatibility use 'Elasticsearch::IndexSettings' instead.</Description>
+        <Description Translatable="1">Number of shards (NS), replicas (NR) and fields limit for the index. Note: 'Elasticsearch::ArticleIndexCreationSettings' is deprecated. For upwards compatibility use 'Elasticsearch::IndexSettings###Default' instead.</Description>
         <Navigation>Core::Elasticsearch::Settings</Navigation>
         <Value>
             <Hash>
--- a/Kernel/Config/Files/XML/Framework.xml
+++ b/Kernel/Config/Files/XML/Framework.xml
@@ -9218,7 +9218,7 @@ via the Preferences button after logging in.
         </Value>
     </Setting>
     <Setting Name="Elasticsearch::IndexSettings###Default" Required="0" Valid="1">
-        <Description Translatable="1">Number of shards (NS), replicas (NR) and fields limit for the indices. This replaces 'Elasticsearch::ArticleIndexCreationSettings' in future versions. If both are present and not equal this one has priority. Use 'Elasticsearch::IndexSettings###...' if you want to define special settings for single indices.'...' may be one of 'Customer', 'CustomerUser', 'Ticket' or 'ConfigItem.'</Description>
+        <Description Translatable="1">Number of shards (NS), replicas (NR) and fields limit for the indices. This replaces 'Elasticsearch::ArticleIndexCreationSettings' in future versions. If both are present and not equal this one has priority. Use 'Elasticsearch::IndexSettings###...' if you want to define special settings for single indices.'...' may be one of 'Customer', 'CustomerUser', 'Ticket' or 'ConfigItem'.</Description>
         <Navigation>Core::Elasticsearch::Settings</Navigation>
         <Value>
             <Hash>
--- a/Kernel/Config/Files/XML/Framework.xml
+++ b/Kernel/Config/Files/XML/Framework.xml
@@ -9206,8 +9206,8 @@ via the Preferences button after logging in.
             </Array>
         </Value>
     </Setting>
-    <Setting Name="Elasticsearch::ArticleIndexCreationSettings###all" Required="0" Valid="1">
-        <Description Translatable="1">Number of shards (NS) and replicas (NR) for the index.</Description>
+    <Setting Name="Elasticsearch::ArticleIndexCreationSettings" Required="0" Valid="1">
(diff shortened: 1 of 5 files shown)


bschmalhofer (commented):
The ` Config   => $ConfigIndexSettings && $ConfigIndexSettings{Ticket} || $Config, ` looks a bit unusual  to me. It's a neat trick but I would add some parens to make obvious what is going on.
` Config   => (($ConfigIndexSettings && $ConfigIndexSettings{Ticket}) || $Config),`

scripts/test/Elasticsearch/IndexSetting.t
```diff
@@ -35,48 +37,57 @@ my $MigObject = $Kernel::OM->Get('Kernel::System::Console::Command::Maint::Elast
...
-$Self->Is(
-    $Index1->{index}->{number_of_shards}, 5,
-    '#1 number_of_shards');
-$Self->Is(
-    $Index1->{index}->{number_of_replicas}, 10,
-    '#1 number_of_replicas');
-$Self->Is(
-    $Index1->{'index.mapping.total_fields.limit'}, 1000,
-    '#1 index.mapping.total_fields.limit');
+Test2::V0::is(
```
bschmalhofer: In test scripts the best practice is to use simply `is()`. E.g.
`is( $index1, { NS => 5, NR => 10, FieldsLimit => 1000 }, 'index1' );`

Besides, I'm confused again. I thought that `_IndexSettingsGet()` should transform the Config to a data structure with keys like `number_of_shards`.

kess-net: That depends on the config settings. As you can see in the expansion tests below in this file it depends on the layout which is set with 'Elasticsearch::IndexSettings'. The first four tests do not use a verbose layout.

Kernel/Config/Files/XML/Framework.xml
```diff
@@ -9217,17 +9217,59 @@ via the Preferences button after logging in.
            </Hash>
         </Value>
     </Setting>
-    <!-- Special settings are possible for the indices Ticket, Customer, CustomerUser and ConfigItem
-         If not set the general settings defined al 'all' are used. -->
-    <Setting Name="Elasticsearch::ArticleIndexCreationSettings###Ticket" Required="0" Valid="1">
-        <Description Translatable="1">Number of shards (NS) and replicas (NR) for the index ticket.</Description>
+    <Setting Name="Elasticsearch::IndexSettings" Required="0" Valid="0">
```
svenoe: Be careful here. This is technically not wrong, but I would prefer, if you used Elasticsearch::IndexSettings###Default. At the moment $ConfigObject->Get('Elasticsearch::IndexSettings') returns the perl structure:
`NS => 1,
NR => 0,
FieldsLimit => 2000,
Ticket => {
    NS => 1,
    ...,
},
Customer => {...},
...,`
While this can be used, as we want people to be flexible with adding more stuff to the index configuration and thus maybe ConfigOptions themselves, too, I would rather have a strict separation there.

Kernel/Config/Files/XML/Framework.xml
```diff
@@ -9206,8 +9206,8 @@ via the Preferences button after logging in.
             </Array>
         </Value>
     </Setting>
-    <Setting Name="Elasticsearch::ArticleIndexCreationSettings###all" Required="0" Valid="1">
-        <Description Translatable="1">Number of shards (NS) and replicas (NR) for the index.</Description>
+    <Setting Name="Elasticsearch::ArticleIndexCreationSettings" Required="0" Valid="1">
```
svenoe, line 9209: Maybe I overlooked this somewhere, but could you please look up, what the FieldsLimit actually does, and whether we really want to add it, please? (This is an addition of Tomohisa and would be new with this patch, too.)

Kernel/System/Console/Command/Maint/Elasticsearch/Migration.pm
```diff
@@ -129,32 +135,32 @@ sub Run {
     if ( $Targets =~ /c/ ) {
         $Self->MigrateCompanies(
             ESObject => $ESObject,
-            Config   => $Config->{Customer} || $Config->{all},
+            Config   => ( ($ConfigIndexSettings && $ConfigIndexSettings->{Customer}) || $Config ),
```
svenoe: I'm a bit torn on this, honestly. I mean on the one hand it's nice, on the other hand, I know that some people will stumble upon this...for now it can stay, I guess, especially as with 10.1 we will sort our the old configuration anyways.

Kernel/System/Console/Command/Maint/Elasticsearch/Migration.pm
```diff
@@ -690,6 +696,9 @@ sub _IndexSettingsGet {
     }
 
     my $IndexSettingsTemplate = $Kernel::OM->Get('Kernel::Config')->Get('Elasticsearch::IndexSettings');
```
svenoe: This is wrong as it is. Generally there is a problem with the naming I was not aware of until now. You are using Elasticsearch::IndexSettings as the new ElasticsearchArticleIndexCreationSettings, which I agree is the better name, as we use it for various indices. In the initial pull request of Tomohisa though, the same name was used for a different setting which is just present in Kernel/Config/Default.pm and can just be changed in Kernel/Config.pm. This is the template for the actual index:

`    $Self->{'Elasticsearch::IndexSettings'} = {
        index => {
            number_of_shards   => '[% Data.NS | uri %]',
            number_of_replicas => '[% Data.NR | uri %]',
        },
        'index.mapping.total_fields.limit' => '[% Data.FieldsLimit | uri %]',
    };`

I would suggest to rename the template to Elasticsearch::IndexTemplate###Default. People should then be able to also put specific templates as Elasticsearch::IndexTemplate###Ticket, or Customer,... to Kernel/Config.pm. (This should be noted in a comment in Default.pm.)

svenoe: Ah, sorry, the formatting here is bad. Please have a look for Elasticsearch::IndexSettings in Kernel/Config/Default.pm.

Kernel/System/Console/Command/Maint/Elasticsearch/Migration.pm
```diff
@@ -734,7 +744,12 @@ sub _ExpandTemplate {
         return;
     }
     elsif ( IsNumber($Node) ) {
-        return $Node;
+        if ( defined( $Config->{ $Param{Key} } ) ) {
+            return $Config->{ $Param{Key} };
+        }
+        else {
+            return $Node;
+        }
```
svenoe, line 753: This whole block is likely not right, just because it is connected to my former comment.

Kernel/Config/Files/XML/Framework.xml
kess-net: ok, point taken.

Kernel/Config/Files/XML/Framework.xml
kess-net: Personally I don't think we need FieldsLimit by default. This defines the maximum number of fields in an index. The default setting is 1000. It is meant to prevent very large mappings and therefore poorly performing searches.
I did understand Tomohisas introduction here as an example for what is possible.
Maybe we should remove it from our configuration and instead deliver an explained example for a more extended configuration.

Kernel/System/Console/Command/Maint/Elasticsearch/Migration.pm
kess-net: Well,... I like compressed code rather than long and verbose blocks of lines reading more like Basic. Sorry if this is really too difficult. I'll try to use a more simple coding in the future.

kess-net: Maybe I should have used the tenary operator?

bschmalhofer: It does look a bit unusual. I probably would have written the check as:
```
my $ConfigIndexSettings = $Kernel::OM->Get('Kernel::Config')->Get('Elasticsearch::IndexSettings') // {};
...
Config   => ($ConfigIndexSettings->{Customer}) // $Config ),
```

kess-net: Ah, yes!  Thank you. I forgot about this operator.

Kernel/System/Console/Command/Maint/Elasticsearch/Migration.pm
kess-net: Did you test this and experienced a wrong behavior - which I would like to see -  or do you just guess?

Kernel/Config/Files/XML/Framework.xml
kess-net: Forget my previous comment. I decided to make usage of the FieldsLimit setting in the installer. I think, Tomohisa intended do this but did not finished it for whatever reason.

kess-net: Forget my previous comment. I decided to make usage of the FieldsLimit setting in the installer. I think, Tomohisa intended do this but did not finished it for whatever reason.

svenoe: Is there anything speaking against setting it to 1000, if it's the default? Probably it doesn't matter much, but it's just keeping everything exactly the same for people who are currently using it.

Issue #1174, opened 2021-08-13
Checkbox Value not correct from Customer Interface


wetzf:
I added a dynamic field of type checkbox to CustomerTicketMessage and hid it with an ACL. With a different ACL I display it if a specific queue is selected. Now if i check the checkbox and submit the form, its value is stored as uncheckes (0). I would expect the value to be checked (1). 
In fact if the field is shown directly and was not hidden via an ACL the Value is stored correctly.
I could reproduce this in Version 10.0.11

wetzf:
It seems like the problem does not occur if i unset "Cache::InBackend" in SysConfig.

stefanhaerter:
I could reproduce the bug. It was due to code clearing the value of fields when they are hidden. I added a special case for the checkboxes because checkboxes work by checked-attribute, not by value. It should be fixed now.

PR #1186, opened 2021-08-18
Issue #1174 Special Handling of Checkboxes for unsetting value when h…

stefanhaerter:
…iding

Closing #1174

### DIFF
--- a/var/httpd/htdocs/js/Core.AJAX.js
+++ b/var/httpd/htdocs/js/Core.AJAX.js
@@ -394,10 +394,10 @@ Core.AJAX = (function (TargetNS) {
                 UpdateTextarea($Element, DataValue);
                 return;
             }
-            
+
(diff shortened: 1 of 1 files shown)


var/httpd/htdocs/js/Core.AJAX.js
```diff
@@ -394,6 +394,12 @@ Core.AJAX = (function (TargetNS) {
                 UpdateTextarea($Element, DataValue);
                 return;
             }
+            
+            // check box elements
+            if ( $Element.is(':checkbox') ) {
+                $Element.prop('checked', false);
```
svenoe: Be careful, in the specific case DataValue will be 0, in which case this behaviour is correct, and I'm not aware that there would ever be a situation in which we will send a 1 here. Still you should at least check it, please. :)

PR #1191, opened 2021-08-23
Dark skin for agent interface

relkai:
At the time of this release, only one default skin is included in OTOBO, and the colors used there are way too bright and shiny for me personally.
So I decided to create a dark skin for OTOBO based on the files included in the default skin.

It's possible that it doesn't work well with all add-on packages, since I focused exclusively on packages I use, which are the following so far:
- FAQ
- Survey


ghost:
Why does no one check this merge request? It would be really helpfull to have multiple skins in Otobo.

Issue #734, opened 2021-01-18
Decide about Kernel::Output::HTML::Notification::AgentSessionLimit


bschmalhofer:
As of OTOBO 10.0.7 the notification Kernel::Output::HTML::Notification::AgentSessionLimit is only a stub. In OTRS this notification is presented when the session limit is approached. As OTOBO has no session limit I would surmise that this notification is not needed in OTOBO. @StefanRother-OTOBO: Is that correct?

When no session limit checks are needed, then it should be decided whether this notification should be kept for just in case, or removed completely.

TODO:

- [x] decide and act accordingly
- [x] fix or remove the TODO test in _scripts/test/Selenium/Agent/Login.t_

svenoe:
I think we can remove it. In the end the code is not lost as it's in git. There is no need for keeping it "just in case".

bschmalhofer:
Ah, I just found out that there still is a session limit in OTOBO. See _Framework.xml_:

```
    <Setting Name="AgentSessionLimit" Required="0" Valid="1">
        <Description Translatable="1">Sets the maximum number of active agents within the timespan defined in SessionMaxIdleTime.</Description>
        <Navigation>Core::Session</Navigation>
        <Value>
            <Item ValueType="String" ValueRegex="^[0-9]{1,8}$">100</Item>
        </Value>
    </Setting>
```
The relevant check is in _Kernel/System/AuthSession.pm_ . Now tthis is a bit inconsistent. When there is a session limit, then a notification that the limit is approaching does make sense. Or should the agent session limit feature be also removed from OTOBO?

bschmalhofer:
Discussed this with @svenoe  and @StefanRother-OTOBO . The conclusion is:

- The agent session check will be kept as currently set up, currently the limit is 100 sessions
- The notification that the limit is approaching will reactivated,

bschmalhofer:
Started to reintegrate the config, tests and implementation of the notification module.

bschmalhofer:
Tests look fine. Closing this issue.

Issue #681, opened 2020-12-27
Clarify status of OTOBOCommunity


bschmalhofer:
In the OTOBO source code and in the test scripts there are some references to OTOBOCommunity. In the original OTRS 6 Community Edition this was OTRSBusiness. 
What is the status of OTOBOCommunity? Can all references be removed?

StefanRother-OTOBO:
Hi Bernhard,

OTOBO has no Community or Business version, I think this is simply a result of the search and replace from the fork.
I'm not sure if it's possible to remove all old OTRSBusiness stuff. Perhaps have a look which test is needed?

I would like to leave the code itself, because I'd like to implement some more business stuff.

@svenoe, what do you think?

Thanks,

Stefan

svenoe:
Yes, this is the renaming of our scripts, quite probably everywhere it occurs. In our first migrations I searched for every instance of OTRS Business and threw out a lot of it, so a lot of what's left should be stuff which I either found interesting, or where I was unsure, and in the end every occurrence should be investigated separately.

We should tidy this up somewhen, but probably after 10.1 is done and we are thinking more about 11.0 features, I guess...

bschmalhofer:
@StefanRother-OTOBO  As far as I see, the directory _Kernel/Config/Files/User_ is used in _Kernel/System/Web/InterfaceAgent.pm_ and in _Kernel/System/SysConfig/Base/OTOBOCommunity.pm_ . So is this, whatever it is, still supported in OTOBO ? Or is it an OTRS Business feature?  In the latter case, I would remove it for the 10.1 release.

bschmalhofer:
@StefanRother-OTOBO  clarified that user configs in  _Kernel/Config/Files/User_ is a rarely used, but potentially useful, feature. It won't be removed. Still, I think that moving the code into a module like _Kernel/System/SysConfig/UserSpecificSettings.pm_ would be a good idea. 
Note that support for user specific config is not implemented in `Kernel::Config` but in `Kernel::System::Web::InterfaceAgent`.

bschmalhofer:
References to OTOBOSTORM and OTOBOCONTROL can be removed as well.

bschmalhofer:
Removed a couple of places where OTOBOCommunity is mentioned. The remaining mention could be removed too, but this is not really required.
[...]

PR #1196, opened 2021-08-26
Issue #734 reinstate session limit notification

bschmalhofer:


bschmalhofer:
For the second, somewhat unrelated, commit see #681,

PR #1221, opened 2021-09-10
Issue #866: make syntax_check.yml work like code_policy.yml

bschmalhofer:
Try whether the changed syntax_check.yml action works.


bschmalhofer:
Not merging, as the branch name was incorrect. 866 should be 856.

PR #1233, opened 2021-09-14
Issue #1132: Rendering dynamic fields shifted to template files

### LINKED ISSUE #625: Fix regular expressions for dynamic fields in CustomerInterface.
Good Morning, 

i have found another error, if a RexEx is entered in a dynamic field of type text, it does not appear in the customer interface.

Thanks!

Stefan

_Originally posted by @StefanRother-OTOBO in https://github.com/RotherOSS/otobo/issues/590#issuecomment-722897929_

### LINKED ISSUE #1132: Think about dynamic field rendering
The dynamic field HTML differs between the customer and agent interfaces. For standard fields this is only small stuff and taken care of via js atm, but at latest with more special fields arriving we should definitely include the interface which the field is rendered for directly in the backend methods, I think. Also we have to think about this with respect to process management, I imagine...

stefanhaerter:
Closing #1132


stefanhaerter:
Last two commits should solve miss-alignment of label for ticket body in CustomerTicketMessage.

Issue #1234, opened 2021-09-15
Allow to activate ElasticSearch in a quick install environment


bschmalhofer:
This bad effect wasn't recognised before, because the unit tests run without ElasticSearch support per default. It should be made easy to set up an environment with ES. Therefore I propose to add a flag `--activate-elasticsearch` to bin\docker\quick_setup.pl. Then test whether there are failures when running with ES.

_Originally posted by @bschmalhofer in https://github.com/RotherOSS/otobo/issues/1192#issuecomment-919964845_

bschmalhofer:
Created https://github.com/RotherOSS/CodePolicy/issues/71 because CodePolicy was naugthy.

PR is merged. Closing this issue.

bschmalhofer:
Reopening because it looks like the SysConfig changes for ElasticSearch were not deployed.

bschmalhofer:
The deployment has been added. Confirmed via the GUI that the deployment was successful. Closing again.

PR #1235, opened 2021-09-15
Issue #1234 quick install

bschmalhofer:


bschmalhofer:
installer.pl should be tested before applying this PR.

bschmalhofer:
Tried the three relevant variants:

- quick_setup.pl with --activate-elasticsearch
- quick_setup.pl without --activate-elasticsearch
- installer.pl, activating ElasticSearch per Default

These variants worked fine. Got the expected errors, in the two cases with active ES, when running _scripts/test/Ticket.t_. Looks good. Merging the PR now.

Issue #1255, opened 2021-09-23
Use caching in the github actions


bschmalhofer:
Here is another idea. When looking at https://github.com/RotherOSS/otobo/runs/3684898399?check_suite_focus=true we see that there is one step that takes most of the time. That's installing the CPAN dependencies from cpanfile.docker. This could be eliminated by using a dedicated Docker image with all the dependencies installed. A better idea might be to simply cache the CPAN dependencies. Either the directory with the installed Perl modules, or the .cpanm workdir.

_Originally posted by @bschmalhofer in https://github.com/RotherOSS/CodePolicy/issues/51#issuecomment-925749171_

bschmalhofer:
It is not clear whether this works as expected. There are several cases to be checked:

- [x] action on push: cache is used when pushing to the same branch
- [x] action on pull_request: cache is used when rerunning the job
- [x] action on pull_request, no cache in parent branch: cache is **not** used on the next PR
- [x] action on pull_request, with cache in parent branch: cache is used on the next PR

bschmalhofer:
Added another action that only builds the cache on the rel-10_1 branch.
Let's not close this issue, until we are sure that no problems crop up.

bschmalhofer:
Another optimisation in cache_local_lib.yml. For getting a checksum on _cpanfile.docker_, we don't need a complete checkout of the branch. We can ask Github directly via the REST interface. Unfortunately the SHA from the webservice is not the same value as returned by `hashfiles()`. This can be redeemed by using `git rev-parse :cpanfile.docker` in the other workflows.

Still not closing, in case problems crop up.

bschmalhofer:
After discussion with @svenoe and consulting #1110 we deciding to merge the two check actions.

bschmalhofer:
Looks good. Only the code_policy.yml workflow is now executed. This workflow contains the syntax check.

bschmalhofer:
This seems to work without hassle. Let's use it for rel-10_0 as well.

bschmalhofer:
Another idea for a speedup. Install Code::TidyAll and dependencies via cpanm. This should make the installation of Debian packages faster.

bschmalhofer:
Closing this issue. Any problems that pop up should be fixed in #1278.

PR #1257, opened 2021-09-23
Issue #1255: add an usage message

bschmalhofer:


bschmalhofer:
Reusing an old branch does not work

PR #1291, opened 2021-10-01
Add nested group support for LDAP user/group sync

fraenki:
Hi,

this is a resubmission of https://github.com/OTRS/otrs/pull/1344, it adds support for nested LDAP groups to otobo. This patch is being used in production for 5 years on OTRS without any issues. Let me know what you think.

An in-depth description of the feature and it's use-case can be found in my original submission:
https://github.com/OTRS/otrs/pull/1344

### DIFF
--- a/Kernel/Config/Defaults.pm
+++ b/Kernel/Config/Defaults.pm
@@ -585,6 +585,9 @@ sub LoadDefaults {
 #        'users',
 #    ];
 
+    # Utilize extended nested group search?
+#    $Self->{'AuthSyncModule::LDAP::NestedGroupSearch'} = '1';
+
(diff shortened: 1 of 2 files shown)


bschmalhofer:
Looks useful and sane to me, but it has been some time since I dealt with LDAP. I propose to include this PR in 10.1.1.
@svenoe , @StefanAbel-OTOBO : What do you think?

svenoe:
I agree. I will bring it up and let @StefanRother-OTOBO  have a look on it the next weeks.

Thanks, @fraenki! :)

svenoe:
Implementing it for 10.1, @StefanRother-OTOBO will do some tests on it.

Thanks again, @fraenki

PR #1310, opened 2021-10-08
Russian translate fixed: 'customers => клиенты' replaced by 'customer…

mdpetrenko:
Russian translate fixed: 'customers => клиенты' replaced by 'customers => компании in order to make 'Customers' and 'Customer users' translate different and more correct

**Old translation:**
Customers => Клиенты
Customer users => Клиенты

**Fixed translation**
Customers => Компании
Customer users => Клиенты

### DIFF
--- a/Kernel/Language/ru.pm
+++ b/Kernel/Language/ru.pm
@@ -444,7 +444,7 @@ sub Data {
         'If nothing is selected, then there are no permissions in this group (tickets will not be available for the customer).' =>
             'Если ничего не выбрано, тогда у клиентов в этой группе не будет прав (заявки будут недоступны клиенту).',
         'Search Results' => 'Результаты поиска',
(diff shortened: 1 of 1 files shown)


svenoe:
Hi Maxim,

thanks for your contribution. For translations however those changes have to be made at - in your case: https://translate.otobo.org/projects/otobo10/otobo/ru/

You can simply search for "Customers" there and if you sort by the fewest words, you directly find it. I didn't put in the change, because I sadly don't speak russian. ;) From my google translation I wonder a bit, whether not maybe "Customer users" should be translated to something more in the direction of "direct user", as "Компании" is translated to "company"
for me, which is a quite specific case of this, and "Клиенты" gets translated to the more general term. Could a "Клиенты" consist of a group of different people, who use OTOBO individually?

I'll leave that to you,... :)

Thanks again, Sven

Issue #1316, opened 2021-10-13
Remove dependency on Plack::Middleware::ForceEnv


bschmalhofer:
The usage of Plack::Middleware::ForceEnv in otobo.psgi looks very wrong. The middleware injects settings in the PSGI variable $Env, not into the system environmen %ENV.     The middleware is also only used for rpc.pl, so any strangeness went propably unnoticed.

bschmalhofer:
While we are at it. the environment variable OTOBO_RUNS_UNDER_PSGI can be eliminated too. There are only two use cases active. 

- SupportDataCollector
- activating DBIx::Connector in Kernel::System::DB

It looks like both cases can be handled with the env variable GATEWAY_INTERFACE.

bschmalhofer:
Plack::Middlware::ForceEnv is removed. Closing this issue.

PR #1323, opened 2021-10-14
Issue #1316 get rid of forceenv

bschmalhofer:


bschmalhofer:
The code policy failures are OK. Looks like the CodePoliy on Githup is not up to date. Kernel::System::Web::Request is declared as a soft dependency and the CodePolicy should honor this.

Issue #1136, opened 2021-07-16
Remove references to CGI modules in otobo.psgi


bschmalhofer:
There is still some support for CGI.pm like objects in OTOBO 10.0.x. But that is mostly useful during testing. _otobo.psgi_ and `Kernel::System::Web::Request` should not explicitly need any modules in the CGI namespace, with the exception `CGI::PSGI`. Let's try to eliminate them.

bschmalhofer:
A special case is the support for `$CGI::POST_MAX` in `Kernel::System::Web::Request`. This variable limits the POST size to the value of `$ConfigObject->Get('WebMaxFileUpload') || 1024 * 1024 * 5`.  It needs to be checked whether this setting is handled be CGI::PSGI in the same way as CGI.pm did in OTOBO 10.0.x without PSGI.

bschmalhofer:
Checked the code and made some small improvements. Test suite still looks fine. Closing this issue.

PR #1383, opened 2021-11-04
Issue #1136 cgi modules

bschmalhofer:


bschmalhofer:
Did some cleanup. Closing the issue only after the next run of the test suite.

Issue #577, opened 2020-10-26
Use DBIx::Connector in the non-PSGI case


bschmalhofer:
Reducing the number of variants is a good thing.
But it has to be pondered whether DBIx::Connector should really be used for console commands and daemon tasks.

bschmalhofer:
Discussed this with @svenoe . There is no good reason to use DBIx::Connector only when running under a webserver. Thus the code can be simplified and DBIx::Connector can be used in all cases.

bschmalhofer:
The test suite looked fine. Closing this issue.

PR #1399, opened 2021-11-11
Issue #577 dbix connector

bschmalhofer:


bschmalhofer:
Ignoring the failure of the CodePolicy check, as the the installed version of the CodePolicy does not yet respect the SoftObjectDependencies setting.

PR #1444, opened 2021-11-24
Feature: ACLs for the Processforms

frankhildebrandt:
Ported our module to be included in core 10.1


svenoe:
Hi Frank,

thanks for the pr, this time I'm not so sure, though. First of all it cannot be integrated into the customer interface just like that, also there are some parts missing on the first look, e.g. dependencies of standard fields (Service -> SLA, etc., which likely results in some bugs in certain cases, I assume, if you are using the ACLPreselectionCache), also I would rather have the convergence block directly in the initial rendering of the mask, and not indirectly activated by triggering a form update from the browser.

I will discuss the priorities with the others and see, whether we implement this functionality the next days.

Best regards, and thanks anyways, Sven

Issue #1434, opened 2021-11-22
Remove unused attributes, like 'Browser',  in Kernel::Output::HTML::Layout


bschmalhofer:
It looks like `Browser` is set but never used. I propose to remove that code.
There are also some attributes that are used only internally, this could be moved into lexical context.

bschmalhofer:
Moved this to Wishlist, as this is not urgent and has some risk.

bschmalhofer:
Removed the branch. Quite sensibly Github keeps the referenced commits, so that they can be cherry-picked in the future.

bschmalhofer:
Reopen this and add ot to the mile stone OTOBO 11.1.    It is just some tidying.

`BrowserMajorVersion` and `BrowserMinorVersion` can also be removed.

bschmalhofer:
The attribute `BrowserJavaScriptSupport` was already removed for rel-11_0 in #2282.

bschmalhofer:
Cherry-picked the changes into rel-11_1. Closing this issue.

PR #1445, opened 2021-11-24
Issue #1434 browser

bschmalhofer:
Hi @svenoe , @stefanhaerter ,

what do you think of this PR? Eliminating unneeded attribtutes in the layout object is safe for core OTOBO, But any package could have used any of the removed attributes. On the other hand, accessing attributes is not part of the documented interface.


svenoe:
grep BrowserJavaScriptSupport Kernel -r # (not safe even for the core...)

We can do this for 11.0, for now I think it is too unimportant and I don't want to risk bugs for stuff like this.

bschmalhofer:
Do this maybe later, maybe in 11.0.1.  Closing the PR without applying.

stefanhaerter:
Remove the wishlist milestone as best practice is that PRs do not have milestones.

PR #1452, opened 2021-11-26
add sysconfig setting to be able to enable the 'exact customer id' se…

reneeb:
…arch field as a default field in the ticket search


svenoe:
Thanks Renee :)

Issue #1470, opened 2021-12-01
Encoding problem for ajax return in the "apache" variant of 10.1


svenoe:
To see the effect visit /otobo/index.pl?Action=AgentStatistics;Subaction=Add and with a German profile click e.g. on "Dynamische Matrix". 

![Screenshot_20211201_120839](https://user-images.githubusercontent.com/48322755/144284680-696c6e64-d60b-42d1-9fae-59d26b4d9ebf.png)

This is currently fetched in line 63 of var/httpd/htdocs/js/Core.Agent.Statistics.js, the probably important side is Kernel/Modules/AgentStatistics.pm Subaction=GeneralSpecificationsWidgetAJAX

bschmalhofer:
First surprise, the Umlaute look as expected to me. Tested with  running Firefox 94.0  and Chromium Version 96.0.4664.45 under Linux. So this effect might be browser dependent. But let's see what possible could go wrong.

The page http://localhost/otobo/index.pl?Action=AgentStatistics;Subaction=Add looks initially fine. The content-type is not surprisingly **ext/html; charset=utf-8**. Clicking on _Dynamische Matrix_ loads the HTML snippet http://localhost/otobo/index.pl?Action=AgentStatistics;Subaction=GeneralSpecificationsWidgetAJAX via the JavaScript code:

                Core.AJAX.FunctionCall(URL, Data, function(Response) {
                    $('#GeneralSpecifications .Content').removeClass('Center').html(Response);
                    Core.UI.InputFields.Activate($('#GeneralSpecifications .Content'));
                    $('#SaveWidget').show();
                }, 'html');
The snippet has content type **text/html**, but the content is UTF-8 encoded, subsequently the output is messed up.

![Bildschirmfoto vom 2021-12-02 11-03-58](https://user-images.githubusercontent.com/46019/144400732-fd28bd56-04bd-4875-b706-d9e1d884b219.png)

Lets doublecheck. Pasting "kÃ¶nnen" into vim and use 'g8' on the funny chars gives us the hex encodings "c373" and "c2b6" for the funny chars. Looking up the code points we confirm that the funny chars are the bytes of UTF-8 encoded 'ö' encoded as UTF-8.

> bes:~/devel/OTOBO/CodePolicy (master)$ uni -x c383
> Ã - U+000C3 - LATIN CAPITAL LETTER A WITH TILDE
> bes:~/devel/OTOBO/CodePolicy (master)$ uni -x c2b6
[...]

PR #1477, opened 2021-12-03
Issue #1470 attachment calls

bschmalhofer:


bschmalhofer:
@svenoe , @stefanhaerter  I think that a second pair of eyes would be a good thing for this issue. Can we discuss the proposed changes?

PR #1498, opened 2021-12-11
add customer ticketcategories service and state

reneeb:
## Proposed change

Enable service and state as possible ticket categories in the customer frontend. Ticket categories are the preferred way to show additional ticket information in the ticket overview.

## Type of change

- '1 - 🚀 feature'           - New feature (which adds functionality to an existing integration)

## Info

- [X] The code change is tested and works locally.(❗)
- [X] There is no commented out code in this PR.(❕)

### DIFF
--- a/Kernel/Config/Files/XML/Ticket.xml
+++ b/Kernel/Config/Files/XML/Ticket.xml
@@ -1925,6 +1925,43 @@
             </Hash>
         </Value>
     </Setting>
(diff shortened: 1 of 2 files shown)


svenoe:
Hi Renee,
yes, we wanted to add more thing here anyways, this is definitely useful.
Thanks, Sven

Issue #1296, opened 2021-10-04
Phrase "otrs" renamed to "otobo" in SysConfig during migration from OTRS to OTOBO


OrionExplorer:
Hello!

I've just performed a successful migration from ((OTRS)) Community Edition 6.0.30 to OTOBO 10.0.12 with migration.pl script.

During standard verification I found an issue with outgoing communication: authentication failed.

Quick look at SendmailModule::AuthUser gave me the answer: username "otrs" has been renamed to "otobo".

I can provide more details if you need.

Regards,
Marcin

bschmalhofer:
Hello Marcin,

this change must have occured in the the migration step OTOBOMigrateConfigFromOTRS. I checked the replacement rules `Kernel::System::MigrateFromOTRS::Base::_ChangeFileInfo()` but found no obvious reason why that renaming would happen. I will test that in my devel installation.

bschmalhofer:
This still happens after fixing #1297, but at least I found the regex that does the replacing. Trying to fix thin in the next days.

bschmalhofer:
While looking into this I noticed that in _Kernel/System/MigrateFromOTRS/CloneDB/Backend.pm_ the following hash is set up:

```
    my %CheckEncodingColumns = (
        'article_data_mime.a_body'              => 1,
        'article_data_mime_attachment.filename' => 1,
    );
```
It looks like this setup has no effect at all. Should be tidied in rel-10_1.

bschmalhofer:
Closed inadvertently.

bschmalhofer:
This funny renaming is happening in the migration step OTOBOMigrateConfigFromOTRS. This step happens after the database tables, including the table _sysconfig_modified_ were transfered. The SysConfig is dumped as a YAML string into the file _SysConfigDump.sysconf_ . Then there are string replacements working on that file. The change file is then reimported in the SysConfig.

The relevant replacements are set up in Kernel::System::MigrateFromOTRS::Base::_ChangeFileInfo()_ . 

In my test migration only two or three value were affected by those string replacements:

> mysql> select name, change_time, convert(effective_value using UTF8) from sysconfig_modified where change_time like '%12-17%';
> +--------------------------+---------------------+------------------------------------------------------------------+
> | name                     | change_time         | convert(effective_value using UTF8)                              |
> +--------------------------+---------------------+------------------------------------------------------------------+
> | SecureMode               | 2021-12-17 15:12:32 | --- '0'
>                                                          |
[...]

PR #1516, opened 2021-12-17
Issue #1296 otrs renamed

bschmalhofer:


bschmalhofer:
Not merging this PR, as the problem is not solved by it,

PR #1533, opened 2022-01-07
one can define more than just the default 9 calendars.

reneeb:
In AdminSLA and AdminQueue this is already implemented, but not in the support data collector


bschmalhofer (approved):
Looks good to me.

svenoe:
Thanks :)

Issue #1339, opened 2021-10-19
Changing a parent queue also changed Changed Date for sub-queue


StefanAbel-OTOBO:
When changing something in a parent queue (), you see in Action=AdminQueue that the sub queue has also been updated - at least the Changed Date adjusts accordingly to the parent Changed Date.

Before you can see the Changed Date 17:00 for Misc::sub:
![image](https://user-images.githubusercontent.com/87359203/137937555-3d2e8287-06d5-4d24-88df-92b811e025ef.png)

After changing something in the parent queue Misc at 17:01 in Action=AdminQueue;Subaction=Change, you see the new Changed Date 17:01 also for Misc::sub:
![image](https://user-images.githubusercontent.com/87359203/137937906-4a3a7a77-35ff-4bbb-a9ef-14622efa9b1e.png)

Tested in 10.0.13 (ov48)

reneeb:
The linked PR avoids updates when the name of the parent queue wasn't updated. There are pro and contra arguments whether a name update of the parent queue should change the "last update" timestamp of the subqueues as it indeed changes the entry in the database. 

The PR can be extended when it is decided that even a name change shouldn't change the "last update" timestamp.

svenoe:
Ah, I should have read the issue before... ;)

Renees pr fixes that subqueues "change" when the name stays the same. Changing the timestamp of subqueues when their name changes is reasonable to me, regardless of whether an agent directly changed it via AdminQueue, or whether it was an automatic change.

PR #1534, opened 2022-01-07
[#1339] Do not update subqueues when the name was not changed

reneeb:
An update of the parent queue did **always** update the names of the subqueues even when the name wasn't changed.

This PR can be extended when a name update shouldn't change the "last updated" timestamp of the subqueues either.

### DIFF
--- a/Kernel/System/Queue.pm
+++ b/Kernel/System/Queue.pm
@@ -1226,6 +1226,8 @@ sub QueueUpdate {
         Type => $Self->{CacheType},
     );
 
+    return 1 if $Param{Name} eq $OldQueue{Name};
+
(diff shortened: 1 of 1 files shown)


svenoe:
Looks good to me, thanks. Concerning the timestamp - I would say if the name is changed it makes sense to change the timestamp. The other cases that I'm aware of, where automatic changes are made, this is done, too, so without checking it further I would guess it is relatively consistent, and I think it makes sense to not only store direct manual changes. (Defining what exactly is counted as "direct" is not so clear mostly, anyways, I guess.)

If you think differently about the timestamps, let's discuss it further.

Issue #1266, opened 2021-09-23
Mandatory DynamicField Dropdown/Multiselect does not work with value "0" 


StefanAbel-OTOBO:
When you have a DF Dropdown e.g. with numbers from 0 to 10 and that DF is a mandatory field (e.g. in AgentTicketPhone or in AgentTicketFreeText), it does not accept the value "0" for the mandatory check and therefore does not save the action or create the ticket.
![image](https://user-images.githubusercontent.com/87359203/134557167-08e83384-1fe6-49c1-810c-62f45ce0c300.png)

![image](https://user-images.githubusercontent.com/87359203/134557240-3ef35508-afe2-4fb7-a5dc-877db93a2761.png)

When you remove the mandatory check and still enter the 0, you see in the ticket history that the value "0" is saved in the ticket - so it doesn't act like it got no value.

![image](https://user-images.githubusercontent.com/87359203/134557004-d68a26fc-d752-48f5-aa07-5c6f614b3b43.png)

This behaviour is the same for DFs from type "Dropdown" and "Multiselect"
Workaround: enter "none" (but you have to live with the mixed up sorting of the values)

reneeb:
I started a PR that should fix this issue...

bschmalhofer:
Added this to the 10.1 milestone as this looks like a significant bug that should be fixed in a maintenance release.

stefanhaerter:
While working at the ACLs, I stumbled across this line:

https://github.com/RotherOSS/otobo/blob/9a6e290ad2a3137576d9940b16026358d8fd9292/Kernel/System/Ticket/FieldRestrictions.pm#L398-L407

I'm not sure if this could potentially be relevant for this issue, leaving this here for further investigation.

stefanhaerter:
I just re-red PR #1535 and compared it with the latest rel-11_0 code because I thought that perhaps we fixed the issue on the way - but the behavior persisted.

stefanhaerter:
Update: The issue is still open. A problem is however, that leaving a field blank yields the value '', so it is not sufficient to check for definedness of the value. The only solution I have at hand is to explicitly check for the value '0', but I am a bit hesitant to do this, to be honest.

svenoe:
Although it would be correct to work with "0"s, too, the cost-benefit ratio is not high enough here, unless someone is willing to pay for it, or a concrete reason comes up, where there is no good workaround at one point.

PR #1535, opened 2022-01-07
[#1266] the number 0 is a pretty valid value for dropdown fields ;-)

reneeb:
Currently the code checks for _true_-ness of the values set. But in Perl
the number 0 is _false_. It should be a valid value though. This commit
changes some checks so that _defined_-ness and length is checked. As 0 is
_defined_ and has a length it is treated as a valid value.

This PR **shouldn't** be merged until the unit tests show that it is safe ;-)

### DIFF
--- a/Kernel/System/DynamicField/Driver/BaseSelect.pm
+++ b/Kernel/System/DynamicField/Driver/BaseSelect.pm
@@ -365,7 +365,7 @@ sub EditFieldValueValidate {
     my $ErrorMessage;
 
     # perform necessary validations
-    if ( $Param{Mandatory} && !$Value ) {
+    if ( $Param{Mandatory} && !( defined $Value && length $Value ) ) {
         return {
             ServerError => 1,
         };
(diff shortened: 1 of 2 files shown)


svenoe:
I played around with it a bit. Actually I thought those routines would test the key of the DynamicField rather than its value, then I wouldn't have seen it as too problematic. I do not really understand why the values are checked, but for this especially you are right, of course. I'm a bit hesitant anyways, as in some cases for example the empty values are added as `'' => '-',`. Because seemingly the values itself are checked it is fine, but I'm not 100% certain if I have the full picture of every eventuality which could arise. Did you run the unit tests with this change being incorporated?

(I would say that even if this introduces some regressions, if they are not huge, we should rather fix those then, and change the logic of the ValueValidate as suggested by you. I just don't want to blindly merge...)

reneeb:
I'll run the unittests at the end of the week. 

The empty value ('') is used to reset (delete the value) of the dynamic field. Adding manually the `'' => '-'` option should have a similar behaviour as the "add empty value" option.

bschmalhofer:
Closing as the associated ticket 1266 is already closed as 'wontfix'.

Issue #1449, opened 2021-11-26
Bug in Action CustomerTicketZoom - when DynamicField from Type Database is used


OlliW:
when a DynamicField from Type Database is used in a Process Action, then occurs an error:

> Traceback (689248):
   Module: Kernel::System::DB::new Line: 135
   Module: Kernel::System::DynamicFieldDB::new Line: 110
   Module: Kernel::System::ObjectManager::_ObjectBuild Line: 323
   Module: Kernel::System::ObjectManager::Get Line: 213
   Module: Kernel::Modules::CustomerDynamicFieldDBSearch::Run Line: 222
   Module: Kernel::System::Web::InterfaceCustomer::Content Line: 1420
   Module: Kernel::System::Web::InterfaceCustomer::Response Line: 1487
   Module: Kernel::System::Web::App::call Line: 86
   Module: Plack::Component::__ANON__ Line: 50
   Module: Plack::App::URLMap::call Line: 71
   Module: Plack::Component::__ANON__ Line: 50
   Module: Plack::Sandbox::_2fopt_2fotobo_2fbin_2fpsgi_2dbin_2fotobo_2epsgi::__ANON__ Line: 199
   Module: Plack::Sandbox::_2fopt_2fotobo_2fbin_2fpsgi_2dbin_2fotobo_2epsgi::__ANON__ Line: 170
   Module: Plack::Sandbox::_2fopt_2fotobo_2fbin_2fpsgi_2dbin_2fotobo_2epsgi::__ANON__ Line: 156
   Module: Plack::Middleware::HTTPExceptions::try {...}  Line: 20

> [Fri Nov 26 09:46:32.643304 2021] [perl:error] [pid 689248] [client x.x.x.x:53983] Can't call method "GetDatabaseFunction" on an undefined value at /opt/otobo/bin/psgi-bin/../../Kernel/System/DynamicFieldDB.pm line 113.\n, referer: http://otobo_101/otobo/customer.pl?Action=CustomerTicketZoom;TicketNumber=2021112698000011

PR #1538, opened 2022-01-10
Issue #1449 new

stefanhaerter:
Closing #1449

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
+++ b/var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
@@ -195,6 +195,12 @@ Core.Agent.DynamicFieldDBSearch = (function(TargetNS) {
 
         if(isJQueryObject($Element)) {
 
+            var LiElementId = $Element.parents().closest('li.Activity').attr('id');
+            var ActivityDialogId = '';
+            if ( typeof LiElementId !== 'undefined' ) {
+                ActivityDialogId = LiElementId.match(/^Process_ActivityDialog-([0-9a-f]{32})$/)[1];
+            }
+
             // Get the ticket id.
             /TicketID=(\d+)/.exec(document.URL);
             TicketID = RegExp.$1;
@@ -209,7 +215,11 @@ Core.Agent.DynamicFieldDBSearch = (function(TargetNS) {
 
             // Register event for the detailed search dialog
             $('#DynamicFieldDBDetailedSearch_' + DynamicFieldName).on('click', function () {
-                OpenDetailedSearchDialog($(this).attr('field'), TicketID);
+                var FieldName = $(this).attr('field');
+                if ( ActivityDialogId !== '' ) {
+                    FieldName = FieldName.substring(0, FieldName.indexOf('_' + ActivityDialogId));
+                }
+                OpenDetailedSearchDialog(FieldName, TicketID);
                 return false;
             });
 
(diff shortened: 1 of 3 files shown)


var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
```diff
@@ -258,13 +258,16 @@ Core.Agent.DynamicFieldDBSearch = (function(TargetNS) {
                     /TicketID=(\d+)/.exec(document.URL);
                     TicketID = RegExp.$1;
 
+                    var liElementId = $Element.parents().closest('li').attr('id');
+                    var activityDialogId = liElementId.match(/^Process_ActivityDialog-([a-f0-9]{32})$/)[1];
```
svenoe: We use capitalized first letters for variables, in js, too.

var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
```diff
@@ -258,13 +258,16 @@ Core.Agent.DynamicFieldDBSearch = (function(TargetNS) {
                     /TicketID=(\d+)/.exec(document.URL);
                     TicketID = RegExp.$1;
 
+                    var liElementId = $Element.parents().closest('li').attr('id');
```
svenoe: It is better to be a bit more precise with the selectors. In this case, e.g. you could use 'li.Activity'. This way you circumvent any strange sideeffects, for new dynamic fields containing lists, e.g.. (Ok, unlikely, but still.)

var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
```diff
@@ -258,13 +258,16 @@ Core.Agent.DynamicFieldDBSearch = (function(TargetNS) {
...
+                    var activityDialogId = liElementId.match(/^Process_ActivityDialog-([a-f0-9]{32})$/)[1];
+
                     // serialize form
                     QueryString = Core.AJAX.SerializeForm($('#'+DynamicFieldName).closest('form'), IgnoreList) + SerializeData(UpdateList);
 
                     QueryString += ";Action="+FrontendInterface;
                     QueryString += ";Term="+encodeURIComponent(Request.term);
                     QueryString += ";MaxResults="+Core.Config.Get('Autocomplete.MaxResultsDisplayed');
-                    QueryString += ";DynamicFieldName="+encodeURIComponent(DynamicFieldName);
+                    QueryString += ";DynamicFieldName="+encodeURIComponent(DynamicFieldName.substring(0, DynamicFieldName.indexOf('_' + activityDialogId)));
```
svenoe: Have you tested whether that works for DB fields which are not part of a process? E.g. in CustomerTicketMessage, or on the agent interface? I would assume a check whether your activityDialogId exists would be good, even if just a warning is thrown.

stefanhaerter: Indeed, this caused an error. Should be fixed with the latest commit

var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
```diff
@@ -258,13 +258,21 @@ Core.Agent.DynamicFieldDBSearch = (function(TargetNS) {
                     /TicketID=(\d+)/.exec(document.URL);
                     TicketID = RegExp.$1;
 
+                    // for usage in process context
+                    var LiElementId = $Element.parents().closest('li.Activity').attr('id');
+                    var DynamicFieldNameQuery = DynamicFieldName;
+                    if (typeof LiElementId !== 'undefined') {
```
svenoe: Did you try that for the normal reply (id=FollowUp) in CustomerTicketZoom? It also has the Activity-class, but the substitution is not needed. Maybe it is necessary to include the "match" as "test" in the if condition.

var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
```diff
@@ -276,7 +284,7 @@ Core.Agent.DynamicFieldDBSearch = (function(TargetNS) {
                         // run the response function to hide the request animation
                         Response({});
                     }
-
+
```
svenoe, line 290: Please remove the blanks. Usually this is done by the CodePolicy, but we have to improve it, to run with js in a meaningful way, too...

stefanhaerter:
Closing this and shifting changes to new branch

stefanhaerter:

Issue #1553, opened 2022-01-23
Fix unit test failures before OTOBO 10.1.1 final release


bschmalhofer:
Here are the current test failures:

```
Test Summary Report
-------------------
/opt/otobo/scripts/test/Compile.t                                                                         (Wstat: 0 Tests: 1787 Failed: 0)
  TODO passed:   1012, 1037
/opt/otobo/scripts/test/DynamicField/EditFieldRender.t                                                    (Wstat: 2048 Tests: 69 Failed: 8)
  Failed tests:  14-21
  Non-zero exit status: 8
/opt/otobo/scripts/test/Selenium/Output/Dashboard/TicketGenericFilter.t                                   (Wstat: 256 Tests: 25 Failed: 1)
  Failed test:  20
  Non-zero exit status: 1
/opt/otobo/scripts/test/Selenium/TestingMethods.t                                                         (Wstat: 0 Tests: 36 Failed: 0)
  TODO passed:   21, 24, 28
/opt/otobo/scripts/test/WebUserAgent.t                                                                    (Wstat: 1024 Tests: 121 Failed: 4)
  Failed tests:  76-77, 89, 101
  Non-zero exit status: 4
Files=961, Tests=82084, 6835 wallclock secs (32.22 usr  3.41 sys + 1758.55 cusr 221.89 csys = 2016.07 CPU)
Result: FAIL
ran tests for product OTOBO 10.1.x on host d9bf5f435b43 .

```

bschmalhofer:
The four failures in _scripts/test/WebUserAgent.t_  are handled in #1250.

bschmalhofer:
The failures in Selenium/Output/Dashboard/TicketGenericFilter.t were due to the new widget showing the last changed tickets. There were some checks that specific tickets should not be shown in the Dashboard after some action in the NewTickets widgets. However the test scripts saw those tickets in the lastchange widget, and was led astray. The fix is to switch to strikter XPath and Jquery selectors that look only at the relevant div.

bschmalhofer:
@stefanhaerter , could you take a look at _scripts/test/DynamicField/EditFieldRender.t_ ? These failures seem to be related to changed HTML in the rendered dynamic fields. See #1132.

stefanhaerter:
Forgot to respond: Should be fixed with PR https://github.com/RotherOSS/otobo/pull/1563

bschmalhofer:
[...]

PR #1563, opened 2022-01-26
Issue #1553 dynamic field bugfixes

stefanhaerter:
Added the concatenation with 'Error' to DivID, like it was done in rel-10_0.

Side notice: I thought about separating DivID and DivIDError for the concatenation, but I am not sure what is the correct and clean way to go in this case so I decided to build it as it was before.

Closing #1553


bschmalhofer (approved):
Looks sensible.

PR #1604, opened 2022-02-08
Issue #625: Display error message for regex fields in Customer Frontend

stefanhaerter:
Closing #625


stefanhaerter:

Issue #1504, opened 2021-12-13
Docker Hub: implement policy concerning the 'latest' tag


bschmalhofer:
Current every new release, regardless whether in 10.0 or on 10.1, sets the lag 'latest'. This does not make sense at all when there is more than on maintained branch.

TODO:

- [x] define a policy which labels should be set
- [x] implement it, propable in _hooks/post_push_

bschmalhofer:
This was discussed between @StefanRother-OTOBO , @svenoe , and @bschmalhofer . The conclusion was:

- the tag 'latest' is deprecated, but will continue to point to the latest patch release of the 10.0 series
- releases that do not match rel-<major>_<minor>_<patch> will get no extra tag, e.g. rel-10_1_Beta2
- patch level releases will get a latest-<major>_<minor> tag

These rules can be implemented in _hooks/post_push_ .

TODO after implementation:

- [x]  sample .env files need to be adapted in docker-compose,  trying to keep rel-10_0 and rel-10_1 with having few differences
- [x] doc-otobo-install might need to be updated
- [x] create a new tag latest-10_0  pointing to the current latest
- [x] backport to rel-10_0

bschmalhofer:
Tested this by setting `DOCKER_TAG` directly in the scripts. Let's see whether a latest-tag will be created after pushing rel-10_1_beta2.

svenoe:
I will close this. If there is a problem, we will see only with the release anyways.

PR #1606, opened 2022-02-08
Issue #1504: switch to tags like: latest-10_1.

bschmalhofer:
Docker tags must match ^rel\-([0-9]+)_([0-9]+)_([0-9]+)$
Keep special case 'latest' for the 10.0 series patch level releases

### DIFF
--- a/hooks/post_push
+++ b/hooks/post_push
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Set additional Docker image tags for automated build on Docker hub
+# Set additional Docker image tags for automated build on Docker hub.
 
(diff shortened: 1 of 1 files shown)


bschmalhofer:
For some reason images on Docker Hub can't easily be tagged remotely. Therefore one needs to do something like:

- `docker pull rotheross/otobo-nginx-kerberos-webproxy:latest`
- `docker tag rotheross/otobo-nginx-kerberos-webproxy:latest rotheross/otobo-nginx-kerberos-webproxy:latest-10_0`
- `docker push rotheross/otobo-nginx-kerberos-webproxy:latest-10_0`

Issue #1608, opened 2022-02-09
Bug in LDAP groups to OTOBO roles synchronization


StefanRother-OTOBO:
I have found a bug in LDAP groups to OTOBO roles synchronization. Once an agent was assigned to roles in OTOBO and then the agent is removed from all Active Directory groups, then the agent keeps the roles rights in OTOBO.

But only when it is removed from all LDAP groups. The reason can be found in line 725 in the Kernel/System/Auth/Sync/LDAP.pm file.

The permissions will not be removed if no more LDAP permissions exist. For me, removing the if query fixed the error:

```
# Rother OSS  / We need to check the permissions also if no more LDAPPermission exists 
    #    if (%RolePermissionsFromLDAP) {
# EO Rother OSS
    
        # get current user roles
        my %UserRoles = $GroupObject->PermissionUserRoleGet(
            UserID => $UserID,
        );

        ROLEID:
        for my $RoleID ( sort keys %SystemRoles ) {

                # if old and new permission for role matches, do nothing
            if (
                ( $UserRoles{$RoleID} && $RolePermissionsFromLDAP{$RoleID} )
                ||
                ( !$UserRoles{$RoleID} && !$RolePermissionsFromLDAP{$RoleID} )
                )
            {
                next ROLEID;
            }

            $Kernel::OM->Get('Kernel::System::Log')->Log(
                Priority => 'notice',
                Message  => "User: '$Param{User}' sync ldap role $SystemRoles{$RoleID}!",
            );

            $GroupObject->PermissionRoleUserAdd(
                UID    => $UserID,
                RID    => $RoleID,
                Active => $RolePermissionsFromLDAP{$RoleID} || 0,
                UserID => 1,
            );
        }
# Rother OSS
#    }  
# EO Rother OSS
```

StefanRother-OTOBO:
Hi Sven,

please can you double check my fix?

Thanks,

Stefan

bschmalhofer:
The fix is sensible as the hash %RolePermissionsFromLDAP can legitimately be empty. Merged a slightly simplified version of the fix. Closing this issue.
See also #1573

svenoe:
There was quite some confusion here...the last commit which is based on the formerly unchanged 10.0 and is then merged to 10.1 should hopefully fix it.

PR #1610, opened 2022-02-11
Issue #1608: handling of %RolePermissionsFromLDAP

bschmalhofer:
is sensible even if the hash is empty


bschmalhofer:
The fix is sansible as the hash `%RolePermissionsFromLDAP` can legitimately be empty. Merged a slightly simplified version of the fix.

Issue #1592, opened 2022-02-02
Remove height in order to see more data than just one line in AgentTicketZoom Processview


StefanRother-OTOBO:
Remove height in order to see more data than just one line in AgentTicketZoom Processview

svenoe:
This needs another look. If we want to keep it that way, at least the class name "CutValue" should be changed if nothing is cut anymore. Maybe though the whole css block as well as the class in the template file can be removed completely as currently only 'display: block' is meaningful.

PR #1626, opened 2022-02-21
Issue #1592: Changed overflow, added word-wrap and max-height

stefanhaerter:
Closing #1592


stefanhaerter:

Issue #282, opened 2020-07-31
Minor improvements to the CustomerInterface design.


svenoe:
- [ ] Make error screens prettier. Kernel/Output/HTML/Templates/Standard/CustomerError.tt (e.g. otobo/customer.pl?Action=CustomerTicketArticleContent), as well as all "grep CenterBox Kernel -r" which affect the customer (e.g. otobo/customer.pl?Action=CustomerTicketZoom;TicketNumber=wrong)
- [x] Improve visualization of readonly fields.
- [ ] Icons: Change checkboxes; replace "x" with icon in select fields.

stefanhaerter:
- [ ] textoverflow: ellipsis for dynamic title fields in TicketInformation in Agent-Interface, implement correct format for CusterInterface
- [ ] Change behaviour for popup widgets, e.g. Print and TicketInformation in CustomerTicketZoom (suggestion: close if focus switches - edit Sven: only close when opening a different popup)
- [x] Selected values in Dropdowns are not colored correctly

svenoe:
I reverted the icon changes for now. We will implement this, but the customer interface needs a little more work.

svenoe:
We close this issue and wait if someone complains about anything here.

bschmalhofer:
Setting the milestone to OTOBO 10.1.16 to indicate the version in which the feature was released.

PR #1654, opened 2022-03-09
Issue #282: Fixing the bug that selected dropdown elements in CustomerInterface are not properly colored

stefanhaerter:
The "background-color: transparent;" prevents the elements from being colored as it is the intended behavior. Though I'm not certainly sure, I could not find any negative impact of deleting this code.

Referencing #282

### DIFF
--- a/var/httpd/htdocs/skins/Customer/default/css/Core.InputFields.css
+++ b/var/httpd/htdocs/skins/Customer/default/css/Core.InputFields.css
@@ -1028,22 +1028,22 @@ div.InputField_ToolbarContainer ul li a:hover {
 }
 
 .jstree-InputField .jstree-wholerow-ul .jstree-wholerow-clicked {
-    background-color: var(--colNotifyOK);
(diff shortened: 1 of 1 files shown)


stefanhaerter:
In order to understand why the other places works and this particular case not: Checking the following places:
- [ ] AgentTicket(Zoom|Email|Process)
- [ ] DynamicField Configurations
- [ ] Customer Screens (Searching where dropdowns are used there)

stefanhaerter:
Investigation suggests that, in case of the Customer Interface, the rule in https://github.com/RotherOSS/otobo/blob/48a8d88698a01373bb4357aefee220526b3f3f16/var/httpd/htdocs/skins/Agent/default/css/Core.InputFields.css#L474 is missing. A possibility would be to add this rule to var/httpd/htdocs/skins/Customer/default/css/Core.InputFields.css, maybe with another color.

stefanhaerter:

PR #1656, opened 2022-03-10
For PUT/PATCH requests, the body is in param PUTDATA/PATCHDATA

reneeb:
If you import the "standard" GenericTicketConnectorREST.yml file and want to update a ticket, one has to use the PATCH HTTP method to send data. The data is in the body, but without this change nothing happens as the HTTP::REST transport class tries to get the query body from the parameter POSTDATA. But that parameter does not exist. Instead the parmeter name depends on the HTTP method used for the request (see https://metacpan.org/dist/CGI/source/lib/CGI.pm#L656). So the HTTP::REST transport class should use those parameter names.


bschmalhofer (approved):
Looks sensible to me. Could you create an issues to which this PR can be attached? So that we can properly track the changes.

Issue #1659, opened 2022-03-11
DynamicFieldDB don't work in processes in the agent interface


reneeb:
When you use a dynamic field of type DB in an activity dialog in the agent interface, it doesn't work. You can search for values, but you can't use them.

The JavaScript code assumes that the activity dialog id is appended to the field name (as it is in the customer interface) and hence it cannot find the field in the activity dialog.

reneeb:
See also https://otobo.de/de/forums/topic/version-10-1-1-dynamische-felder-db/

PR #1660, opened 2022-03-11
Issue #1659: Strip the activity dialog id from the fieldname only whe…

reneeb:
…n it is present.

Otherwise the field name becomes '' and that leads to error messages in the system log ('Need DynamicFieldName') and
to JavaScript errors as it cannot find the field in the dialog.

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
+++ b/var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
@@ -605,8 +605,9 @@ Core.Agent.DynamicFieldDBSearch = (function(TargetNS) {
         }
 
         var FieldNameLong = Field;
-        if ( ActivityDialogID != '' ) {
-            Field = Field.substr(0, Field.indexOf('_' + ActivityDialogID));
+        var IndexOfActivityDialogID = Field.substr(0, Field.indexOf('_' + ActivityDialogID));
+        if ( ActivityDialogID != '' && IndexOfActivityDialogID > 0 ) {
+            Field = Field.substr(0, IndexOfActivityDialogID);
         }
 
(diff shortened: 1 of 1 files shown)


var/httpd/htdocs/js/Core.Agent.DynamicFieldDBSearch.js
```diff
@@ -605,8 +605,9 @@ Core.Agent.DynamicFieldDBSearch = (function(TargetNS) {
         }
 
         var FieldNameLong = Field;
-        if ( ActivityDialogID != '' ) {
-            Field = Field.substr(0, Field.indexOf('_' + ActivityDialogID));
+        var IndexOfActivityDialogID = Field.substr(0, Field.indexOf('_' + ActivityDialogID));
+        if ( ActivityDialogID != '' && IndexOfActivityDialogID > 0 ) {
```
stefanhaerter, line 609: I would suggest to shorten it a bit, e.g. like:

if ( ActivityDialogID != '' && Field.indexOf('_' + ActivityDialogID > 0) ) {

However, this assumes that the original DynamicFieldName has a length > 0, but as far as I know, a DynamicField is required to have a name.

reneeb: I used an extra variable as the result is used twice (see new line 610).

I don't get the last pragraph....

stefanhaerter: The FieldName here is '[DynamicFieldName]\_[ActivityDialogID]', and if the DynamicFieldName would be empty, 0 would be a valid index for '_' + ActivityDialogID. But I can't think of a case where this would happen, so it looks ok to me. @svenoe What do you think?

Issue #1661, opened 2022-03-11
DynamicFieldDB: Changing the value triggers warnings


reneeb:
If you define a dynamic field of type database and that field stores string values in the otobo database, changing the value for the dynamic field triggers warnings like:

```
Argument "root@localhost" isn't numeric in sort at /opt/otobo/bin/psgi-bin/../../Kernel/System/DynamicField/Driver/BaseDatabase.pm line 71.
```

The data are handled as numerics in the `ValueIsDifferent` method of _Kernel::System::DynamicField::Driver::BaseDatabase_

reneeb:
A sample dynamic field:

![image](https://user-images.githubusercontent.com/144096/157826750-90821b35-8641-428f-a7fe-552729db1599.png)

Use that field in AgentTicketFreeText. Assign a value and then change the value...

PR #1662, opened 2022-03-11
Issue #1661: Do not sort arrays numerically as the dynamic field can …

reneeb:
…store string values

and that would trigger a warning like 'Argument 'root@localhost' isn't numeric in sort at /opt/otobo/bin/psgi-bin/../../Kernel/System/DynamicField/Driver/BaseDatabase.pm line 71.'


svenoe:
I thought this was already included a month ago or so...maybe a different place.

PR #1671, opened 2022-03-14
Issue #1659 index correction

stefanhaerter:
IndexOfActivityDialog holds the substring, which leads the expression IndexOfActivityDialogID > 0 to return false, so the name is not shortened.

Closing #1659


stefanhaerter:

Issue #1673, opened 2022-03-15
Limit Dev::UnitTest::Run to a specific package


bschmalhofer:
For running the tests of an OTOBO package it would nice if the console command `Dev::UnitTest::Run` were enhanced. Pass the _,sopm_ file and run only the _.t_ files that are mentionend in the Filelist.

_Originally posted by @bschmalhofer in https://github.com/RotherOSS/FAQ/issues/7#issuecomment-1068142288_

svenoe:
I agree. Also I think this could be done within 10.1, as it only affects the testing environment, it's a simple thing, and the benefit is quite high.

bschmalhofer:
Sounds good, I have moved the milestone to 10.1.2.

bschmalhofer:
The console command `Dev::UnitTest::Run` uses the SysConfig setting _UnitTest::DefaultDirectory._

  `  my $DefaultDirectory = $Kernel::OM->Get('Kernel::Config')->Get('UnitTest::DefaultDirectory');
`
However this setting is nowhere declared nor documented. Let's remove it in rel-11_0.

bschmalhofer:
In order to be consistent the files from the .sopm files must be treated like whitelist patterns. Running Dev::UnitTest::Run with passing a .sopm file worked at least once. Thus closing this issue.

bschmalhofer:
It is already possible to run only the tests mentioned in an .sopm file. But this could be made more useful if one could specify the name of an already installed  OTOBO package. The information should be available as the .sopm file is stored in the database table `package_repository`.

PR #1677, opened 2022-03-16
Issue #1673 test only sopm

bschmalhofer:


bschmalhofer:
Github currently has some problems running the CodePolicy. But the policy check was fine locally. Thus merging this PR.

PR #1771, opened 2022-05-12
Fix typo

abrain:


GritRother-OTOBO:
Hi abrain, 

Thanks for this!

As our source language is English, changes to the German text have to be done in Weblate which is where the translation is maintained (=> [translate.otobo.org](url)).

I have corrected the German text there, and will close this issue.

Thanks, Grit

Issue #1810, opened 2022-06-06
Unittest failure in Selenium/Customer/DynamicFields/DynamicFieldDatabase.t


bschmalhofer:
This is a new failure. Should be fixed for the next patch level release.

bschmalhofer:
This might be a sporadic failure. I ran  this as a single test script and it worked.

stefanhaerter:
Do you have further information about the failure?

bschmalhofer:
Here is the log output from two of my latest test runs. Not that the exception occurred at different locations.

[issue_1810.log](https://github.com/RotherOSS/otobo/files/8861261/issue_1810.log)

stefanhaerter:
I've seen this problem earlier but couldn't reproduce it reliably. It occurs while testing the DynamicField of type Database within the following steps:

- Searching an element via the standard input field and clicking it so that it is added as a result element
- Searching an element via detailed search and clicking it in the resulting table so that it is added
- Verifying that the added elements are actually present in the interface

Here the error occurs since the ResultElement2, which was added via detailed search, cannot be found. At first glance, I think this might be a timing issue or something related.

PR #1822, opened 2022-06-17
Issue #1810: wait until element exists

daxim:


bschmalhofer (commented):
Looks good. way better than all those unmotivated sleeps-

Issue #1826, opened 2022-06-27
Switch ticket age to create date in ticket lists


GritRother-OTOBO:
It would be great if the ticket age would switch to the ticket creation date in all ticket lists (customer and agent interface) after e.g. 24 hours for easier readability.

stefanhaerter:
Note for Development: 
https://github.com/RotherOSS/otobo/blob/rel-10_1/Kernel/Output/HTML/TicketOverview/Small.pm#L531
If-Else-Block with Kernel::System::DateTime->Delta (assuming that all tickets have their creation date in the past)

stefanhaerter:
I suppose that PR #2197 and #1105 target different areas of this issue.

stefanhaerter:
Additional suggestion: Change title attribute of TicketAge to hold create date. Doing so, you could easily see the ticket create date while hovering on the ticket age.

For comparison: This is the way GitHub has implemented it.

stefanhaerter:
The actual implementation has significantly diverged from the original feature request and undergone several reconceptions. This should be taken into account if the feature is to ever be included into core.

stefanhaerter:
PR merged, closing the issue.

PR #1833, opened 2022-07-05
Issue #1826: Added if-else-block for different age formatting for tic…

stefanhaerter:
…kets older than 24h

Closing #1826

### DIFF
--- a/Kernel/Output/HTML/TicketOverview/Medium.pm
+++ b/Kernel/Output/HTML/TicketOverview/Medium.pm
@@ -398,10 +398,25 @@ sub _Show {
     );
 
     # create human age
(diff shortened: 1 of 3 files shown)


svenoe (commented):
Please check whether the change can be shifted to CustomerAge.
Also please merge into rel-11

stefanhaerter:
As far as I understand sub CustomerAge, the implementation cannot be shifted because CustmerAge only has access to the age and not to the date.

stefanhaerter:
After discussion, the decision is to add an optional parameter 'CreatedAt' to CustomerAge function as well as a sysconfig option and rendering the age as date if the parameter is present and the sysconfig option is set.

stefanhaerter:
Closing this in favor of #2197

Issue #1341, opened 2021-10-20
"This feature is currently not available." when adding a calendar appointment.


StefanAbel-OTOBO:
When adding a new appointment to any calendar, you still see the rest of an "advertisement" for the OTRS Feature AddOn "Calendar Resource Planning":

Resource
This feature is currently not available.

![image](https://user-images.githubusercontent.com/87359203/138069443-c9fb01d8-0168-458b-8ef3-f74ce3e28038.png)

Either we should implement a resource planning to fill that gap or we should remove that part.

Tested with OTOBO 10.0.13 (ov48).

stefanhaerter:
Merged.

PR #1835, opened 2022-07-06
Issue #1341: Removed "This feature is currently not available" from t…

stefanhaerter:
…emplate.

Closing #1341

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AgentAppointmentEdit.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentAppointmentEdit.tt
@@ -111,8 +111,8 @@
                 [% Data.ResourceIDStrg %]
 [% END %]
             </div>
-[% END %]
         </fieldset>
+[% END %]
         <fieldset class="TableLike">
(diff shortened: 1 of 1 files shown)


Kernel/Output/HTML/Templates/Standard/AgentAppointmentEdit.tt
```diff
@@ -111,12 +111,6 @@
                 [% Data.ResourceIDStrg %]
 [% END %]
             </div>
-[% ELSE %]
-            <div class="Field">
-                <p class="FieldExplanation">
-                    [% Translate("This feature is currently not available.") | html %]
-                </p>
-            </div>
 [% END %]
```
svenoe: Doesn't this belong below </fieldset> now?
Edit: I meant the "[% END %]"

stefanhaerter: Because of Stefan's description, I thought that this snippet was not wanted anymore - do we want to keep it?

Issue #1531, opened 2022-01-06
Handle erroneously configured TreeView DynamicFields.


svenoe:
Select Fields in TreeView do not work if sub categories are set without the base category being present.
(E.g. plant::tree and plant::bush being used without plant being used. This is not always the case - see ov71 for an example.)

The nicest way to handle this, would probably be to add the node dynamically but disable its selection. Else at least a useful error message should appear somewhere.

stefanhaerter:
Besides the changes I already made, it seems to be necessary to alter the code in https://github.com/RotherOSS/otobo/blob/rel-10_1/Kernel/Output/HTML/Layout.pm#L5938 in the following way:

`$Row->{Key}      = '-';` to `$Row->{Key}      = substr $Row->{Key}, 0, (length($Row->{Key}) - length('_Disabled'));`

because in Core.UI.TreeSelection.js the options are sorted alphabetically and the Key '-' causes the items to be displayed wrongly, as far as I see. However, I'm not entirely sure if the Key currently used is needed in any other place.

stefanhaerter:
Closed with merging #2529

PR #1841, opened 2022-07-14
Issue #1531: Passing Sort param "TreeView" in case treeview is config…

stefanhaerter:
…ured

Closing #1531


stefanhaerter:
Seems that a sorting bug is still remaining.

stefanhaerter:

Issue #1447, opened 2021-11-25
Disabled dynamic fields shown in CustomerTicketZoom


cugu:
Hi,

The CustomerTicketZoom view shows the fields "process" and "activity" even when they are disabled in the screens.

reneeb:
Process and Activity are special as they are always shown in process tickets:

see lines 1713 .. 
https://github.com/RotherOSS/otobo/blob/163c96b00a62d38e5509d3e595c69e256537a51d/Kernel/Modules/CustomerTicketZoom.pm#L1713

@svenoe This could be configurable as those information can "leak" internal organisational data to customers...

svenoe:
I agree. I have not looked into where exactly the info is shown, but if CustomerTicketZoom###DynamicField can be used here, in a sensible way, we should do that, instead of introducing new options. I don't know whether we have two separate places, though. (Maybe we can just disable one, then.)

reneeb:
I'll take a closer look at it at the end of the week.

stefanhaerter:
So far it appears to be the case that, if Process and Activity are enabled vor CustomerTicketZoom, each appears twice in the information widget. This is due to the fact that they are rendered once via the dynamic fields handling and once via the code marked above. So I would suggest to remove the code fragment in Kernel/Modules/CustomerTicketZoom.p, L.1713ff, but before it has to be checked if the handling in Kernel/System/DynamicField/Driver/ProvessManagement/ActivityID.pm and ProcessID.pm is correct.

Another problem has occured to me while investigating this issue. When reaching https://github.com/RotherOSS/otobo/blob/rel-10_1/Kernel/System/ProcessManagement/Activity.pm#L100, some activity entities appear to be broken, like the second item here:

```
$Activity = {
              "Activity-991d30fdf276f4344bc5b608c24c1b60" => {
                                                               ActivityDialog => { 1 => "ActivityDialog-a66dedc3b5f7213496d52550c0ebef93" },
                                                               ChangeTime => "2022-07-06 06:59:28",
                                                               CreateTime => "2022-07-06 06:59:28",
                                                               ID => 1,
                                                               Name => "TestAktivit\xE4t1",
                                                             },     
              "Activity-e2712e2c4433f53e745b48d94b09a2c3" => {
                                                               1 => "ActivityDialog-967df513b3b04e8e89b0777c0bcc3313",
                                                               ActivityDialog => {}, 
                                                             },     
            }    
```  

[...]

PR #1845, opened 2022-07-19
Issue #1447 customerticketzoom

stefanhaerter:
1) I removed the code for displaying process and activity permanently, so from now on, if this information is wanted, the corresponding dynamic fields have to be used
2) A reference bug caused broken activity information in the config at runtime

Closing #1447

### DIFF
--- a/Kernel/Modules/AgentTicketZoom.pm
+++ b/Kernel/Modules/AgentTicketZoom.pm
@@ -1563,7 +1563,7 @@ sub MaskAgentZoom {
 
         # get next activity dialogs
         my $NextActivityDialogs;
-        if ( $Ticket{$ActivityEntityIDField} ) {
+        if ( $Ticket{$ActivityEntityIDField} && IsHashRefWithData( $ActivityData ) && IsHashRefWithData( $ActivityData->{ActivityDialog} ) ) {
             $NextActivityDialogs = ${ActivityData}->{ActivityDialog} || {};
         }
         my $ActivityName = $ActivityData->{Name};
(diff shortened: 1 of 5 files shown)


bschmalhofer (approved):
Looks fine. The removed code avoids the problem of inadvertently changing the data in the Kernel::Config instance. 
I have one suggestion for further code cleanup. In line 1782 the variable $NextActivityDialogs is assigned a new value, unrelated to the previous value. In those cases I tend to give the variable a new name, declared with my.

stefanhaerter:
Thank you for the good suggestion. I would make this change, but this seems to happen in L1795 also. I don't know if that makes much of a difference, but intuitively I would want to avoid initializing several variables which are not used later on - what do you think?

Maybe it could be a solution not to overwrite the existing variable but to actually filter the values with map?

bschmalhofer:
I haven't really looked closely at the code. The reuse of variables in different contexts is IMHO a red flag. If it occurs in several lines then it might make sense to rename or remove those variables. But of course only when there is no danger of unwanted side effects.

stefanhaerter:
I tried to reduce the assignments with my latest commit. Do you think this looks ok?

svenoe (commented):
Please confirm that $ActivityData and $ActivityData->{ActivityDialog} are always hashrefs.
Please look for LayoutObject->Block(ProcessData) in AgentTicketZoom.

stefanhaerter:
I added checks with IsHashRefWithData on ActivityData and ActivityData->{ActivityDialog}.
AgentTicketZoom doesn't render a ProcessData block and apparently never did since the initial OTOBO commit.

stefanhaerter:
Closing this PR as work continued on a different branch.

PR #1861, opened 2022-08-01
Update TicketCreate.pm

StefanAbel-OTOBO:
changed the wrong default value for "TimeUnit", as it was the one from "Responsible"


bschmalhofer (approved):
15 seems to be a sensible default value, as the passed value ends up in time_accounting.time_unit   which seems to be accounted time in seconds.

Issue #1722, opened 2022-04-19
Update JavaScript libraries


bschmalhofer:
Bring the JS libraries to their current versions. Be inspired by https://github.com/znuny/Znuny/blob/rel-6_3_1/CHANGES.md.

- jQuery
  - upgrade JQuery to 3.6.0
  - keep the _.min_ in the name _jquery-3.6.0/jquery.min.js_
- remove jquery-migrate-3.0.0 as it was unused and not up to date
- jquery-ui
  - generate a custom download with the feature list from https://github.com/OTRS/otrs/commit/05dbcf42415d3db54822486128f9169baea0f450
  - Note that Znuny has expanded the feature list for no apparent reason
  - Also update jquery-ui.min.css, apparently the CSS hasn't been updated since 2014
  - jquery-ui.min.css already includes jquery-ui.structure.min.css and jquery-ui.theme.min.css  See https://jqueryui.com/upgrade-guide/1.11/#simplified-custom-and-quick-downloads
  - update the icons in skins/Customer/default/css/thirdparty/ui-theme/images/
  - update the icons in skins/Agents/default/css/thirdparty/ui-theme/images/
  - remove the images that were no longer included in the custom download

bschmalhofer:
The current version of CKEditor 4 is 4.18. But upgrading CKEditor should be handled in #790.

bschmalhofer:
Looks like _jquery-browser-detection.js_ can't be removed as the method `browser` is used in _./var/httpd/htdocs/js/Core.App.js_ .

bschmalhofer:
Also update qunit to version 2.18.2

But as of 2024-06-27  qunit is now at 2.21.0

bschmalhofer:
The test suite ran fine. PR is merged. Let's keep this issue open, checking whether problems crop up.

bschmalhofer:
Two more updates as seen in https://github.com/znuny/Znuny/tree/rel-6_3-dev/var/httpd/htdocs/js/thirdparty:
- momentjs-2.29.1   (actually using 2.29.3)
- nunjucks-3.2.2

stefanhaerter:
I updated the jquery-ui from 1.12.1 to 1.13.2. First, I adjusted the paths. Afterwards, I committed the js and css files. 

The jquery-ui homepage appears to offer different themes. An error in customer user selection seemed to occur from a wrong theme. As far as I could manage to find out, the right one is **Smotheness**.

[...]

PR #1864, opened 2022-08-10
Issue #1722 jquery UI

stefanhaerter:
See comment in #1722  for details.


bschmalhofer (commented):
Updating to the current version of JQuery UI is fine IMHO. @svenoe might recall whether there were any problems when Jquery UI  was updated for OTOBO 10.0.16.
It might be better to merge first with rel-10_0 and then merge rel-10_0 into rel-10_1. But I can't pinpoint an reason for that gut feeling.

svenoe:
Via merge

Issue #1735, opened 2022-04-25
Investigate error message when ACL with empty ConfigMatch is imported


stefanhaerter:
When an ACL with empty ConfigMatch is either imported or set within a Selenium Test, the following error can be observed via fred:

Use of uninitialized value in concatenation (.) or string at Kernel/System/ACL/DB/ACL.pm line 546.

This seems to be due to [this line](https://github.com/RotherOSS/otobo/blob/6accae067cf02e02a6fa259a7ecd6b4fd8253e51/Kernel/System/ACL/DB/ACL.pm#L553). $Param{ConfigMatch} is undef, so the 'eq' seems to fail.

stefanhaerter:
The purpose at this point is to fetch the ACL data from the database and compare it to the input data for verifying that it has been saved correctly. Taking into account L.468-490, ConfigMatch and ConfigChange are allowed to be either a Hashref with data or undefined. For evading the undefined-case. I would suggest to use the $ConfigMatch and $ConfigChange variables defined in L.465f instead of $Param{ConfigMatch} and $Param{ConfigChange}:

if ( 
        $CurrentName
        && $CurrentName eq $Param{Name}
        && $CurrentComment eq $Param{Comment}
        && $CurrentDescription eq $Param{Description}
        && $CurrentStopAfterMatch eq $Param{StopAfterMatch}
        && $CurrentValidID eq $Param{ValidID}
        && $CurrentConfigMatch eq (defined $Param{ConfigMatch} ? $Param{ConfigMatch} : $ConfigMatch)
        && $CurrentConfigChange eq (defined $Param{ConfigChange} ? $Param{ConfigChange} : $ConfigChange)
        )    
    {    
        return 1;
    }

bschmalhofer:
It's not only for evading the undefined case. Lets consider the case where a ConfigMatch already is in the database and a ConfigMatch had been passed to the ACLUpdate() method. Then a String would be compared against a hash reference, which does not make sense. So I second to to use the $ConfigMatch and $ConfigChange variables  for the comparison.

stefanhaerter:
Merged.

PR #1870, opened 2022-08-12
Issue #1735: Prevent use of undefined value in string eq

stefanhaerter:
Closing #1735


stefanhaerter:
I have to rethink this.

PR #1871, opened 2022-08-12
Issue #1735: Use predefined variables for config match and config cha…

stefanhaerter:
…nge instead of param values

Closing #1735


bschmalhofer (commented):
Looks sensible to me. 
Could you check whether adding a   $Param{StopAfterMatch} //= 0    makes sense too? Then StopAfterMatch would behave just like Description.

stefanhaerter:
As far as I tested it, exporting an ACL produces a yaml with 'StopAfterMatch: 0'. I wonder if there is a way to produce a config with an undef value or an empty string if not on purpose. However, I think this is a good addition.

bschmalhofer (approved):
Looks fine to me.

PR #1893, opened 2022-08-30
Fix typo on customer logo

niccord:
The width had a typo on the value of _oooNavigation_ and _oooSignet_ classes, this will fix it.
This bug can also be found in version 10.0.x


bschmalhofer:
Thanks Nikola!

PR #1935, opened 2022-09-12
Fix package upgrade when its name changes case

### LINKED ISSUE #1931: Support for package names that differ only in case
Hi all,
when upgrading a package, if it has changed the case of some letters in its name, then the upgrading process will be stuck in a loop.

For example:
upgrading my package from `OpmTest v.10.0.0` to `opmtest v.10.0.1`
this is the output in the console:
``` bash
$ bin/otobo.Console.pl Admin::Package::Install opmtest-10.0.1.opm
Installing package...
Deep recursion on subroutine "Kernel::System::Package::PackageInstall" at /opt/otobo/Kernel/System/Package.pm line 834.
Deep recursion on subroutine "Kernel::System::Package::PackageUpgrade" at /opt/otobo/Kernel/System/Package.pm line 519.
```

Since I think the uniqueness of a package name should not be based on the case of its letters, I think you should change the check on `Kernel::System::Package`.

Cheers!

niccord:
fix #1931

Fix package upgrade when its name changes case

### DIFF
--- a/Kernel/System/Package.pm
+++ b/Kernel/System/Package.pm
@@ -816,7 +816,7 @@ sub PackageUpgrade {
     my $InstalledVersion = 0;
     for my $Package ( $Self->RepositoryList() ) {
 
-        if ( $Structure{Name}->{Content} eq $Package->{Name}->{Content} ) {
+        if ( lc $Structure{Name}->{Content} eq lc $Package->{Name}->{Content} ) {
 
             if ( $Package->{Status} =~ /^installed$/i ) {
                 $Installed          = 1;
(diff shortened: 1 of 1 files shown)


svenoe:
Hi Nicola,
thanks, at the place I'm against implementing the change, as it is inconsistent, though. Package name changes are not allowed, automatically. You can use
`<PackageMerge Name="OldPackageName" TargetVersion="10.1.x"></PackageMerge>`
within the sopm to replace packages with different name(s) with the new one during installation.
Please write again, if I'm missing something, or misunderstanding the problem. For now I will close this.
Best regards, Sven

Issue #1938, opened 2022-09-15
Print company ticket in CustomerTicketOverview shows wrong "printed by" information


StefanRother-OTOBO:
Hi,

When I print a ticket in the OTOBO Customer Area -> Company Tickets, the customer who created the company ticket is displayed at the top under "printed by", not the customer who is printing the ticket.

Thanks,

Stefan

bschmalhofer:
Merged the recent changes in rel-10_1 mainly because the CodePolicy action failed. I suppose this was because the branch was still using an old version of _cpanfile.docker_. Thus an outdated set of CPAN libraries was used, where LibUtil.c could not be linked any more.

stefanhaerter:
Fix merged and looking good. Closing the issue.

PR #1943, opened 2022-09-16
Issue #1938: If in customer interface, use UserID of initiating user

stefanhaerter:
Closing #1938


svenoe (commented):
In my opinion this is not the correct way to fix things. Just because CustomerUserB prints the ticket of CustomerUserA, still the CustomerData of the ticket should be the data of CustomerA. If this is used nowhere else, but just in the "printed by" field, then there would be no wrong effects, but I would assume this is not the case. I think the "printed by" has to be solved more locally (without having have looked at it, though).

Kernel/Output/PDF/Ticket.pm
```diff
@@ -173,11 +173,8 @@ sub GeneratePDF {
...
     my $CustomerUserObject = $Kernel::OM->Get('Kernel::System::CustomerUser');
     my %CustomerData;
-    if ( $Interface{Customer} ) {
-        %CustomerData = $CustomerUserObject->CustomerUserDataGet(
-            User => $Param{UserID},
-        );
-    }
+    my %PrintingCustomerData;
+
     elsif ( $Ticket{CustomerUserID} ) {
```
bschmalhofer: Shouldn't this be an 'if' instead of an 'elsif' ?

stefanhaerter: Yes, it should - strange, I thought I pushed the correction.

Kernel/Output/PDF/Ticket.pm
```diff
@@ -184,6 +186,17 @@ sub GeneratePDF {
         );
     }
 
+    if ( $Interface{Customer} ) {
+        %PrintingCustomerData = $CustomerUserObject->CustomerUserDataGet(
+            User => $Param{UserID},
+        );
+    }
+
+    if ( !defined %PrintingCustomerData )
```
bschmalhofer: Checking the definedness of hashes and arrays is not really supported. The documentation says:
            Use of "defined" on aggregates (hashes and arrays) is no longer
            supported. It used to report whether memory for that aggregate
            had ever been allocated. You should instead use a simple test
            for size:

                if (@an_array) { print "has array elements\n" }
                if (%a_hash)   { print "has hash members\n"   }

Kernel/Output/PDF/Ticket.pm
```diff
@@ -184,6 +186,17 @@ sub GeneratePDF {
         );
     }
 
+    if ( $Interface{Customer} ) {
+        %PrintingCustomerData = $CustomerUserObject->CustomerUserDataGet(
+            User => $Param{UserID},
+        );
+    }
+
+    if ( !%PrintingCustomerData )
```
bschmalhofer: Looks better. Did you run the CodePolicy on that code? I think that Perl::Critic would put the opening brace at the end of line 195.

stefanhaerter: CodePolicy didn't change anything - made this change by myself.

Kernel/Output/PDF/Ticket.pm
```diff
@@ -258,9 +258,7 @@ sub GeneratePDF {
         'DateFormat',
     );
     if ( $Interface{Agent} ) {
-        $PrintedBy .= ' ' . $TicketInfo{User}{UserFullname} . ' ('
-            . $TicketInfo{User}{UserEmail} . ')'
-            . ', ' . $Time;
+        $PrintedBy .= " $TicketInfo{User}{UserFullname} ($TicketInfo{User}{UserEmail}), $Time";
```
bschmalhofer, line 261: I would always use the long form with nested hashes, but I'm not sure whether that convention has been agreed upon:
`$TicketInfo{User}{UserFullname}  => $TicketInfo{User}->{UserFullname} `

stefanhaerter: As discussed, I would leave this as it is, regarding the pre-existing code style.

bschmalhofer (approved):
No objections from my side.

Issue #1946, opened 2022-09-19
Error message in Log when changing data for a customer user with no password


stefanhaerter:
It is possible to create a customer user without setting a password. If the user data is altered later, the following error message appears in the log:
Use of uninitialized value $CurrentUserData{"UserPassword"} in string ne at /opt/otobo/bin/psgi-bin/../../Kernel/Modules/AdminCustomerUser.pm line 374.

bschmalhofer:
@stefanhaerter Why is this labelled as invalid? That a warning is emitted with in a valid use case is IMHO a bug.

stefanhaerter:
You're right. I wasn't sure about how to label this. Since the process works, I thought the description "Something isn't right" would fit best.

bschmalhofer:
Yes, the description of the label was confusing. I changed it to "The report was either incorrect, or it describes the correct behavior."

stefanhaerter:
Slightly related to #779

stefanhaerter:
Closed with merging #2556

PR #1947, opened 2022-09-19
Issue #1946: Add check for customer user password

stefanhaerter:
Closing #1946

### DIFF
--- a/Kernel/Modules/AdminCustomerUser.pm
+++ b/Kernel/Modules/AdminCustomerUser.pm
@@ -371,7 +371,7 @@ sub Run {
                 );
             }
 
-            if ( $CurrentUserData{UserPassword} ne $GetParam{UserPassword} ) {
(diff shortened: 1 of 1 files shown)


svenoe (changes requested):
This would mean that passwords cannot be deleted.

stefanhaerter:
Ah, I forgot this case. if (defined ...) should fix this, right?

Kernel/Modules/AdminCustomerUser.pm
```diff
@@ -371,6 +371,7 @@ sub Run {
                 );
             }
 
+            $CurrentUserData{UserPassword} //= '';
             if ( $CurrentUserData{UserPassword} ne $GetParam{UserPassword} ) {
```
svenoe: Please don't change the variable, instead use:
if ( ( undef // "123" ) eq "123" ) {print "klappt\n"}

stefanhaerter:
Closing in favor of #2554

PR #1958, opened 2022-09-29
Fixed hidden filters for Medium and Preview

tipue-dev:
It is possible to set filters in the Small view. Those filters are still used for Medium and Preview, but here is the delete button missing, so you can't remove the filters or even see, that filters are set.
image
This PR fixes this by adding the stored filters to $Self which results in usable delete buttons.


bschmalhofer:
Hi Tim,
thanks for fixing this bug. I merged the PR and the fix will be available in OTOBO 10.1.6. 
Additionally @daxim enhanced the Selenium test script.

Best regards,
   Bernhard

PR #1959, opened 2022-09-29
Fixed double encoding for dynamic field values in the linked-tickets-table

tipue-dev:
The values of dynamic fields in the complex link table in the AgentTicketZoom are overly encoded. `<b>Big</b>` Is the value
image

This change would fix it to

image


stefanhaerter (approved):
I investigated this and it looks like a useful change to me. I tested to introduce js injection with HTMLOutput => 0, but it didn't work.

svenoe:
Better late than never... thanks

Issue #1573, opened 2022-01-26
Tidying of Kernel::System::Auth::Sync::LDAP


bschmalhofer:
No special reason, but it is good to be consistent. `_ConvertFrom()` is also a method.

Maybe also check whether the `$FindMember` as an anonymous sub is really the best solution.

bschmalhofer:
See also #1291 and https://github.com/OTRS/otrs/pull/1344.

bschmalhofer:
It also would be very nice to have better test coverage of LDAP related features. AFAIK, in OTOBO there is not even a mock LDAP server that could be used for running tests against.

bschmalhofer:
Merged the easy parts of the open patch.

bschmalhofer:
Provided an updated version of the patch as a PR.

bschmalhofer:
Hi Frank, @fraenki ,

when working on a simplified PR I noticed the call to `Net::LDAP::search()` in https://github.com/RotherOSS/otobo/blob/rel-10_1/Kernel/System/Auth/Sync/LDAP.pm#L937. Do you know whether the parameter `Attributes`is correct? In the [documentation](https://metacpan.org/dist/perl-ldap/view/lib/Net/LDAP.pod#search-(-OPTIONS-)) I have only seen the parameter `attrs`.

Best regards,
    Bernhard

fraenki:
> Do you know whether the parameter Attributes is correct? In the [documentation](https://metacpan.org/dist/perl-ldap/view/lib/Net/LDAP.pod#search-(-OPTIONS-)) I have only seen the parameter attrs.

It's been so long since I've touched this code, so I don't know. :(
But after reading the documentation I agree with your assessment, it should most likely be `attrs`.

bschmalhofer:
Hi Frank, @fraenki, 
after looking at the code of [Net::LDAP](https://metacpan.org/dist/perl-ldap/view/lib/Net/LDAP.pm) I concluded that the parameter `Attributes` is silently ignored. That is, only the parameter `attrs` would be part of the message sent to the LDAP server.  as I don't want to change the current behavior I went for simply removing `Attributes`, 

The current behaviour is that the default attributes of the found LDAP entries are returned. In principle it would make sense to request that only the needed attributes, even if they are not declared as default, are included. But that would be a different issue.

Could you give the newest version,
[...]

Issue #1811, opened 2022-06-06
Log spam: Exiting subroutine via last at...


fraenki:
When using nested LDAP groups in Otobo 10.1.4, the log is spammed with debug information on every logon:

```
[Mon Jun  6 17:51:21 2022] -e: Exiting eval via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[Mon Jun  6 17:51:21 2022] -e: Exiting subroutine via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[Mon Jun  6 17:51:21 2022] -e: Exiting subroutine via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[Mon Jun  6 17:51:21 2022] -e: Exiting subroutine via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[Mon Jun  6 17:51:21 2022] -e: Exiting subroutine via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[Mon Jun  6 17:51:21 2022] -e: Exiting subroutine via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[Mon Jun  6 17:51:21 2022] -e: Exiting subroutine via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[Mon Jun  6 17:51:21 2022] -e: Exiting subroutine via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[Mon Jun  6 17:51:21 2022] -e: Exiting subroutine via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[Mon Jun  6 17:51:21 2022] -e: Exiting subroutine via last at /opt/otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 963.
[...]

PR #1969, opened 2022-10-10
Issue #1573 ldap tidying take3

bschmalhofer:


fraenki:
As [discussed here](https://github.com/RotherOSS/otobo/issues/1811#issuecomment-1274574212) I've tested the patch today. Unfortunately it results in a `500 Internal Server Error` and these are the errors from the log files:

```
[Mon Oct 31 17:47:27 2022] -e: Use of uninitialized value $GroupDN in hash element at otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 908.

[Mon Oct 31 17:47:27.346502 2022] [perl:error] Can't use an undefined value as a SCALAR reference at otobo-10.1.4/bin/psgi-bin/../../Kernel/System/Auth/Sync/LDAP.pm line 938.\n, referer: https://example.com/otobo/index.pl
```

I've tested it against Otobo 10.1.4, if that matters.

bschmalhofer:
Hi Frank,
I have looked at the PR again and was a bit overwhelmed. I still see the added clarity when having _FindMembers() as a regular subroutine. I'll provide a new PR. Closing this PR.

PR #1989, opened 2022-10-21
Allowed for Redis with protected mode and password

escix:
Allowed for Redis with protected mode and password


bschmalhofer:
Hi @escix,
 I see that there are are a lot of changes from rel10_1 in your PR. I have therefore merged rel-10_1 into rel-11_0. Could you resubmit your PR based on the updated rel-11_0. The PR  should be much smaller now.

Best regards,
   Bernhard

PR #1990, opened 2022-10-21
Rel-11_0

escix:
Added password config for Redis::Fast


bschmalhofer:
Hi @escix,

I think that we need to take a step back here. Could you open first a Github Issue that states what problem you are solving? We will then see whether a fix should become part of the next patch release in the 10.1.x series or be included in OTOBO 11.0.x. 
There is also a contributers guide at https://doc.otobo.org/manual/developer/10.1/en/content/contributing/sending-contributions.html . This guide is still a bit rudimentary, improving it is on our long TODO list.

Best regards,
   Bernhard

bschmalhofer (commented):
This is merging changes done for 11.0.x back to 10.1.x. Backporting fixes is possible, but should be kept to a minimum.

escix:
Hi @bschmalhofer,
Unfortunately I am not a developer and know very little about GIT processes.
I noticed that when the Redis is configured in protected mode with password, OTOBO is unable to connect to the database, there is also no option to include the password in the config.
Considering this is a security issue I modified my code to include the password in the Redis.pm file and also uploaded to GIT.
Due to time constraint I am unable to proceed further, hopefully someone will merge this in after a good review.
Thanks for all your help for keeping this software updated, I understand the scale of work involved.

bschmalhofer:
Hi @escix ,
thanks for the feedback. That clarifies it. I have created the issue https://github.com/RotherOSS/otobo/issues/1991 as a wishlist item. So it is undetemined when this will be worked on.

Best regards,
   Bernhard

Issue #2002, opened 2022-11-04
Remove obsolete support for setting Frontend::NavBarOutputModule'


bschmalhofer:
Looks like there is obsolete code in _./Kernel/Output/HTML/Layout.pm_
```
bes:~/devel/OTOBO/otobo (rel-10_1)$ grep -r NavBarOutputModule .
./Kernel/Output/HTML/Layout.pm:    my $NavBarOutputModuleConfig = $ConfigObject->Get('Frontend::NavBarOutputModule');
./Kernel/Output/HTML/Layout.pm:    if ( ref $NavBarOutputModuleConfig eq 'HASH' ) {
./Kernel/Output/HTML/Layout.pm:        my %Jobs = $NavBarOutputModuleConfig->%*;

```
Looks like this was added in 2012 but was never used or documented:
```
bes:~/devel/znuny/Znuny (rel-6_4-dev)$ git blame ./Kernel/Output/HTML/Layout.pm | grep NavBarOutputModule
Verarbeite Zeilen: 100% (6625/6625), fertig.
90f323a2541 Kernel/Output/HTML/Layout.pm  (Martin Gruner        2015-06-02 09:47:57 +0200 3295)     my $NavBarOutputModuleConfig = $ConfigObject->Get('Frontend::NavBarOutputModule');
a9e5df3ceb0 Kernel/Output/HTML/Layout.pm  (Carlos Rodriguez     2012-08-21 21:25:31 +0000 3296)     if ( ref $NavBarOutputModuleConfig eq 'HASH' ) {
a9e5df3ceb0 Kernel/Output/HTML/Layout.pm  (Carlos Rodriguez     2012-08-21 21:25:31 +0000 3297)         my %Jobs = %{$NavBarOutputModuleConfig};

```

bschmalhofer:
@svenoe pointed out that `Frontend::NavBarOutputModule` could be used by OTOBO packages for adding to the navigation bar. So let's not remove the code, only add a code comment.

bschmalhofer:
Added a code comment. Closing the issue as invalid.

PR #2003, opened 2022-11-04
Issue #2002: Frontend::NavBarOutputModule is obsolete

bschmalhofer:
Hallo @StefanRother-OTOBO ,

just to be sure, do you see any use for `Frontend::NavBarOutputModule`? I didn't find any mention of that setting.


bschmalhofer:
Decided to keep the code, as output modules could be used by OTOBO packages.

Issue #2016, opened 2022-11-17
Please change default state type to "pending reminder" when adding a new state


StefanAbel-OTOBO:
When you add a new state in Action=AdminState;Subaction=Add, there is a state type pre-selected - depending on the language, as it is the first possible state type alphabetically.
It would be great, if the default state type could be "pending reminder", as this the state type you select most often.

stefanhaerter:
Can it be guaranteed that the state type "pending reminder" is present on every system? Or could the state type be deleted somehow?

StefanAbel-OTOBO:
probably in the database, but I am not aware of any way how to delete a state type in the frontend somehow

svenoe:
We won't change this in OTOBO 11.0 anymore. Probably needs a quick Abstimmung, again, also...

svenoe:
We should change this for consistency reasons between languages. It was decided to add an empty value, though, to force a concious selection. (Note it is 11.1, not 11.0!)

stefanhaerter:
PR merged, closing the issue.

PR #2017, opened 2022-11-18
Issue #2016: Set default selected state type to pending reminder

stefanhaerter:
Closing #2016

### DIFF
--- a/Kernel/Modules/AdminState.pm
+++ b/Kernel/Modules/AdminState.pm
@@ -352,10 +352,11 @@ sub _Edit {
         SelectedID => $Param{ValidID} || $ValidListReverse{valid},
         Class      => 'Modernize Validate_Required ' . ( $Param{Errors}->{'ValidIDInvalid'} || '' ),
     );
+    # default selected state type is 'pending reminder'
     $Param{StateTypeOption} = $LayoutObject->BuildSelection(
         Data       => { $StateObject->StateTypeList( UserID => 1 ), },
         Name       => 'TypeID',
-        SelectedID => $Param{TypeID},
+        SelectedID => $Param{TypeID} || 4,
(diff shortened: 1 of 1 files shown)


bschmalhofer (approved):
Usually, magic number like 4 should be avoided, but I don't really have a better idea. Approving as the magic number is at least documented in a code comment.

stefanhaerter:
I had the same thought. Only alternative I'm aware of would be to use Selected => 'pending reminder', but I would prefer the id over the name, and as @StefanAbel-OTOBO indicated [here](https://github.com/RotherOSS/otobo/issues/2016#issuecomment-1319951988), there is no known sane way to delete the state type, so I hope this is an acceptable option.

Kernel/Modules/AdminState.pm
```diff
@@ -352,10 +352,11 @@ sub _Edit {
         SelectedID => $Param{ValidID} || $ValidListReverse{valid},
         Class      => 'Modernize Validate_Required ' . ( $Param{Errors}->{'ValidIDInvalid'} || '' ),
     );
+    # default selected state type is 'pending reminder'
     $Param{StateTypeOption} = $LayoutObject->BuildSelection(
         Data       => { $StateObject->StateTypeList( UserID => 1 ), },
         Name       => 'TypeID',
-        SelectedID => $Param{TypeID},
+        SelectedID => $Param{TypeID} || 4,
```
svenoe: Please use Kernel::System::State->StateTypeLookup() for 'pending reminder' and just set the default ID, if an ID is found.

stefanhaerter:
Closing the PR as decision was made to implement this differently and for 11.1.

Leaving the branch alive for the moment to look up details later on.

Issue #1984, opened 2022-10-19
Dynamic Field Database search backend produces undefined data due to off by one error


stefanhaerter:
Code line:
https://github.com/RotherOSS/otobo/blob/d89a619725a6450d18a317f3df2dad70931d4d42/Kernel/System/DynamicFieldDB.pm#L560

Because $RowCount is initialized with scalar @Row, which gives the number of elements in the array, instead of $#Row, which gives the highest index, the returned data contains '{ Data => undef, Label => undef },'.

The same initialization is used in the function DatabaseSearchDetails, but there a prevention against empty items is included. I would opt for fixing this at both locations nonetheless.

bschmalhofer:
At least in the first case you could also write something like  `for my $Item ( @Row ) { .. `.

stefanhaerter:
Good suggestion, I will check this. Maybe I was tempted by doing as few changes as possible ;)

stefanhaerter:
PR merged.

bschmalhofer:
Reopen because there was a failure in _scripts/test/DynamicField/DynamicFieldDB.t_

stefanhaerter:
Test looking fine so far. @bschmalhofer can you confirm?

bschmalhofer:
Running the test suite with Docker looks fine. Closing this issue again.

PR #2048, opened 2022-12-19
Issue #1984: adapt test script to the fixed bug

bschmalhofer:
A bogus search result is no longer returned. This is expected as an incorrect loop was fixed in the module code.
Use Test2::V0 more consistently, for a more useful output of the differences in the data structures.


stefanhaerter (approved):
Looks Good To Me.

PR #2059, opened 2022-12-21
Issue #1881: Updating backup.pl to prevent own column charset and use…

### LINKED ISSUE #1881: Handle charset that are declared per column
Having the charset utf8 declared on the individual columns is an interesting case.

First, let's sanity check that charset utf8 vs. utf8mb4 is really the problem. The problematic emoji is indeed encoded in four bytes:

> $ uni -8 😊
> 😊 - U+1F60A - F0 9F 98 8A - SMILING FACE WITH SMILING EYES

So, this theory is confirmed.

I assume that you used the streamlined migration method for migrating the data from OTRS to OTOBO:
https://doc.otobo.org/manual/installation/10.0/en/content/migration-from-otrs-6.html#optional-step-streamlined-migration-of-the-database .
Checking the script _scripts/backup.pl_, I found that there is an oversight. The charset on the tables is adapted and the COLLATE on columns is removed in the subroutine `MySQLBackupForMigrateFromOTRS()`, see https://github.com/RotherOSS/otobo/blob/rel-10_1/scripts/backup.pl#L578. But the `CHARACTER SET utf8` is not removed. This oversight seems to be the root cause of the problem.

Now for the fix. The statement `alter table article_data_mime convert to character set utf8mb4 collate utf8mb4_general_ci;` is likely not enough, as the charset set of the columns have higher priority than the default charset of the table. So the relevant statement is more like: 

> # For each column:
> ALTER TABLE table_name CHANGE column_name column_name VARCHAR(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
See https://mathiasbynens.be/notes/mysql-utf8mb4 .

My preferred solution would be to redo the migration. Either with a fixed script _scripts/backup.pl_ or with using the standard migration method.

_Originally posted by @bschmalhofer in https://github.com/RotherOSS/otobo/issues/1879#issuecomment-1222036258_

dennis-dko:
… it from the table now.

### DIFF
--- a/scripts/backup.pl
+++ b/scripts/backup.pl
@@ -577,6 +577,7 @@ END_MESSAGE
 
             # substitutions for changing the character set
             $Line =~ s/DEFAULT CHARSET=utf8/DEFAULT CHARSET=utf8mb4/;    # for CREATE TABLE
+            $Line =~ s/CHARACTER SET .*?\s//;                            # for CREATE COLUMN
             $Line =~ s/utf8mb4mb4/utf8mb4/;                              # in case it already was utf8mb4
             $Line =~ s/utf8mb3mb4/utf8mb4/;                              # in case of some mixup
             $Line =~ s/utf8mb4mb3/utf8mb4/;                              # in case of some mixup


scripts/backup.pl
```diff
@@ -577,6 +577,7 @@ sub MySQLBackupForMigrateFromOTRS {
 
             # substitutions for changing the character set
             $Line =~ s/DEFAULT CHARSET=utf8/DEFAULT CHARSET=utf8mb4/;    # for CREATE TABLE
+            $Line =~ s/CHARACTER SET .*?\s//;                            # for CREATE COLUMN
```
bschmalhofer, line 580: I propose to substitute with a single space, just in case there is only a single whitespace character after the character set.
Is it assured that there is whitespace after the charset, e.g. when it is the last column?

bschmalhofer: The questions are resolved. The trailing whitespace can be removed, as there is white space before 'CHARACTER' .

bschmalhofer (approved):

PR #2062, opened 2022-12-22
Issue #1908 is null is not added when extra source columns are migrated

### LINKED ISSUE #1908: 'IS NULL'  and default value are not added when extra source columns are migrated
_migration.pl_ supports the feature that extra columns from the source database are also migrated to the target database. These extra columns should have the same name and the same data type as the source columns. But it looks like this does not work for 'IS NULL'. The reason is that in e.g. in https://github.com/RotherOSS/otobo/blob/rel-10_1/Kernel/System/MigrateFromOTRS/CloneDB/Driver/mysql.pm#L296 the check is `$ColumnInfos{IS_NULLABLE} =~ /no/` while IS_NULLABLE is either **NO** or **YES**.

dennis-dko:


Kernel/System/MigrateFromOTRS/CloneDB/Driver/mysql.pm
```diff
@@ -288,15 +289,20 @@ sub AlterTableAddColumn {
     my $SQL         = qq{ALTER TABLE $Param{Table} ADD $Param{Column} $ColumnInfos{DATA_TYPE}};
 
     if ( $ColumnInfos{LENGTH} ) {
-        $SQL .= " \($ColumnInfos{LENGTH}\)";
+        $SQL .= " ($ColumnInfos{LENGTH})";
     }
```
bschmalhofer, line 294: I think that $ColumnInfos{COLUMN_DEFAULT} should be quoted, just in case it contains a single quote character.

PR #2079, opened 2023-01-04
Issue #2065: Fix problem with lowercase tables in migration script to…

### LINKED ISSUE #2065: Tablename in uppercase breaks the migration script
_migration.pl_ uses the **ListTables()** function that makes all tablenames to lowercase. The Problem is that the _migration.pl_ use a `"SELECT COUNT(*) FROM <table_name>" `to check all these tables, if the real tablename is in uppercase like **Persons** instead of **persons** the check won't success and the migration script doesn't allow you to continue.

dennis-dko:
… check if it is exist.

### DIFF
--- a/Kernel/System/DB.pm
+++ b/Kernel/System/DB.pm
@@ -899,7 +899,7 @@ The table names are lower cased.
 =cut
 
 sub ListTables {
-    my $Self = shift;
+    my ( $Self, %Param ) = @_;
 
     my $SQL = $Self->GetDatabaseFunction('ListTables');
 
@@ -920,7 +920,11 @@ sub ListTables {
 
     my @Tables;
     while ( my ($Table) = $Self->FetchrowArray() ) {
-        push @Tables, lc $Table;
+        if ( $Param{Original} ) {
+            push @Tables, $Table;
+        } else {
+            push @Tables, lc $Table;
+        }
     }
 
     return @Tables;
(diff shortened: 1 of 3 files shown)


Kernel/System/DB.pm
```diff
@@ -920,7 +920,11 @@ sub ListTables {
 
     my @Tables;
     while ( my ($Table) = $Self->FetchrowArray() ) {
-        push @Tables, lc $Table;
+        if ( $Param{Original} ) {
+            push @Tables, $Table;
+        } else {
```
bschmalhofer: The coding standard is the 'else' is on a new line. Could you check whether CodePolicy automatically fixes this?

Kernel/System/DB.pm
```diff
@@ -899,7 +899,7 @@ The table names are lower cased.
 =cut
 
 sub ListTables {
-    my $Self = shift;
+    my ( $Self, %Param ) = @_;
```
bschmalhofer, line 902: The new parameter should be documented in the POD.

bschmalhofer (commented):
I'm not happy with yet another list of tables. There is even an reason for that. Let's say an OTRS package adds a table 'dummy1' and for some reason we have only have a table called 'Dummy1' in the OTRS database. Then there will be no error from the sanity check, but maybe a problem when the migration proceeds.
But then I don't really have a good idea how to decide whether a table is relevant or not. I don't really want to add a dependency on the OTOBO DB  just for checking the OTRS DB. 

By now I tend to like the first approach, where for the SanityCheck there is no lower casing. Then the SantiyCheck will only complain about actually broken tabels, which can be considered to be a good thing.

Issue #1948, opened 2022-09-19
Make output of Main::Config::Dump more useful


bschmalhofer:
For Complex setting the output of `Maint::Config::Dump` is not really useful.

> $ bin/otobo.Console.pl Maint::Config::Dump Loader::Module::AgentTicketProcess
> 002-ProcessManagement=HASH(0x562abcf20c10);

I propose to emit the full data structure as JSON per default.

See also https://otobo.de/en/forums/topic/help-example-request-for-console-pl-maintconfigdump/.

bschmalhofer:
Tentatively changes the output to JSON. Here are some some examples:

```
> bes:~/devel/OTOBO/otobo (issue-#1948-json_dump)$ ./bin/otobo.Console.pl Maint::Config::Dump CurrentDeploymentID
> "2"
> bes:~/devel/OTOBO/otobo (issue-#1948-json_dump)$ ./bin/otobo.Console.pl Maint::Config::Dump ACLKeysLevel1Change
> {
>    "Possible" : "Possible",
>    "PossibleAdd" : "PossibleAdd",
>    "PossibleNot" : "PossibleNot"
> }
> bes:~/devel/OTOBO/otobo (issue-#1948-json_dump)$ ./bin/otobo.Console.pl Maint::Config::Dump dummy
> Error: The config setting dummy could not be found.
> bes:~/devel/OTOBO/otobo (issue-#1948-json_dump)$ ./bin/otobo.Console.pl Maint::Config::Dump PerformanceLog::File
> "/opt/otobo/var/log/Performance.log"
> bes:~/devel/OTOBO/otobo (issue-#1948-json_dump)$ ./bin/otobo.Console.pl Maint::Config::Dump Package::RepositoryRoot
> [
>    "https://ftp.otobo.org/pub/otobo/misc/packages/repository.xml"
> ]
> bes:~/devel/OTOBO/otobo (issue-#1948-json_dump)$ ./bin/otobo.Console.pl Maint::Config::Dump Loader::Agent::CommonCSS
> {
>    "000-DynamicFieldDB" : [
>       "DynamicFieldDB.css"
>    ],
>    "000-Framework" : [
>       "Core.Reset.css",
>       "Core.Default.css",
>       "Core.Header.css",
>       "Core.OverviewControl.css",
>       "Core.OverviewSmall.css",
>       "Core.OverviewMedium.css",
>       "Core.OverviewLarge.css",
>       "Core.Footer.css",
[...]

PR #2124, opened 2023-01-17
Issue #1948 config dump as json

bschmalhofer:


bschmalhofer:
Applied the PR event though CodePolicy complained. The violation was a harmless Miscellanea::ProhibitUselessNoCritic. The 'no critic' was considered to be useless as the CodyPolicy in Github differs from the CodePolicy used in the development of rel-11_0.

Issue #2177, opened 2023-02-18
set 'no critic' for Modules::ProhibitEvilModules when using Data::Dumper


bschmalhofer:
Handle the fallout from the changed CodePolicy, see  https://github.com/RotherOSS/CodePolicy/issues/104 .

bschmalhofer:
The CodePolicy results, using HEAD of the branch master,  look fine for all files in rel-11_0. Closing this issue.

PR #2178, opened 2023-02-18
Issue #2177: TidyAll::Plugin::OTOBO::Perl::Dumper is gone

bschmalhofer:
replaced by the Perl::Critic policy Modules::ProhibitEvilModules


bschmalhofer:
The CodePolicy were from Data::Dumper, because the CodePolicy in github is not up to date. Merging anyways.

Issue #2179, opened 2023-02-18
Make creation of test driver object optional in test scripts


bschmalhofer:
Using the package `Kernel::System::UnitTest::Driver` adds little value for test scripts that are using mostly `Test2::V0`. Yet an instances of that class is created when `Kernel::System::UnitTest::RegisterDriver;` is loaded. It would be nice to not having to call `use Kernel::System::UnitTest::RegisterDriver;` just for creating the `$Kernel::OM` object.
 
Implementation could be done by moving the import sub `Kernel::System::UnitTest::RegisterDriver::import();` to a dedicated module `Kernel::System::UnitTest::RegisterOM;`. Care must be taken that existing test scripts do not have to be updated. This can be achievied  by  `Kernel::System::UnitTest::RegisterDriver` depending on  `Kernel::System::UnitTest::RegisterOM`.

bschmalhofer:
See also https://github.com/RotherOSS/otobo/issues/87.

bschmalhofer:
The test suite looks fine. PR is merged. Closing this issue.

bschmalhofer:
Reopening because it is confusing that a package called `Kernel::System::UnitTest::RegisterOM` suddenly deals with calling `Kernel::System::UnitTest::Helper::DESTROY()`.

bschmalhofer:
Closing again, after moving the END block to a more fitting module.

PR #2180, opened 2023-02-19
Issue #2179 register om

bschmalhofer:


bschmalhofer:
The three CodePolicy violations are due to the outdated CodePolicy used in the Github action. Thus merging despite the violations.

Issue #2213, opened 2023-03-05
Test failure in OverviewSmall.t


bschmalhofer:
Recently I started to see test failures for _scripts/test/Selenium/Output/Ticket/OverviewSmall.t_. This was for OTOBO rel-10_0 and rel-11_0.

Here are the latest runs:
- prove_11.0.x_2023-02-19-151541_rel-11_0.out: OK
- prove_10.1.x_2023-02-20-102943_rel-10_1.out:   OK
- prove_11.0.x_2023-02-21-162557_rel-11_0.out: OK
- prove_11.0.x_2023-02-28-184717_issue-#88-sax_parser.out: not OK
- prove_11.0.x_2023-03-01-203449_issue-#88-xml_simple.out: not OK
- prove_10.0.x_2023-03-04-140324_rel-10_0.out : not OK
- prove_11.0.x_2023-03-05-143255_issue-#2211-https_on.out: not OK 

```

# Failed test 'RunTest: no exception'
# at /opt/otobo/Kernel/System/UnitTest/Selenium.pm line 306.
# Exception: WaitFor(JavaScript => return typeof($) === 'function' && $('#ColumnFilterDynamicField_DFtest12604043456000000120-TicketNew option[value="Key1%261"]').length;) timed out at /opt/otobo/scripts/test/Selenium/Output/Ticket/OverviewSmall.t line 249.
# Looks like you failed 1 test of 39.
[15:24:42] /opt/otobo/scripts/test/Selenium/Output/Ticket/OverviewSmall.t .............................................. 

```
Needs to be investigated.

stefanhaerter:
Failure is related to fix for #2143.

bschmalhofer:
Tests look fine again for rel-11_0. Closing this issue.
```
Test Summary Report
-------------------
/opt/otobo/scripts/test/Compile.t                                                                         (Wstat: 0 Tests: 1774 Failed: 0)
  TODO passed:   1137, 1234
/opt/otobo/scripts/test/Selenium/TestingMethods.t                                                         (Wstat: 0 Tests: 36 Failed: 0)
  TODO passed:   21, 24, 28
Files=959, Tests=80826, 5632 wallclock secs (27.58 usr  3.23 sys + 1568.54 cusr 221.08 csys = 1820.43 CPU)
Result: PASS
ran tests for product OTOBO 11.0.x on host 4e50a7f55912 .

```

PR #2214, opened 2023-03-06
Issue #2213: Dynamic field value in column filtering is no longer dou…

stefanhaerter:
…ble encoded, adjusting the unit test accordingly.

Fix for #2213


bschmalhofer (approved):
Looks good to me. Extra points for removing more code than adding.

Issue #2127, opened 2023-01-18
Expose Plack::Request::headers() in the ParamObject


bschmalhofer:
Sometimes OTOBO needs to fiddle with the request headers. E.g. in _Kernel/GenericInterface/Transport/HTTP/REST.pm_:
```
        my %RequestHeaders;
        for my $EnvKey ( sort $ParamObject->HTTP() ) {
            my $HeaderKey = substr $EnvKey, 5;    # remove leading HTTP_
            $HeaderKey =~ s{_}{-}xmsg;
            $RequestHeaders{$HeaderKey} = $ParamObject->HTTP($EnvKey);
        }

```
In these cases it might be nice to have an instance of [HTTP::Headers](https://metacpan.org/pod/HTTP::Headers). This could be achieved by adding a wrapper method around `Plack::Request::headers()`.

bschmalhofer:
Replacing the calls to HTTP() in OTOBO core. In the test suite there were more lines printed than usually. This seems to be related to Elasticsearch running out of disk space. Otherwise the test suite looks fine. The PR is merged, closing this issue.

PR #2218, opened 2023-03-08
Issue #2127 http headers fast take 2

bschmalhofer:


bschmalhofer:
Complaints from CodePolicy are due to differences in the different versions of the CodePolicy. Thus ignoring the violations and merging anyways.

Issue #1777, opened 2022-05-13
Resolve unit test failures when syncing with S3 is enabled


bschmalhofer:
Currently there are still 23 test scripts that are failing when syncing with MinIO is activated. These need to be acivated.

One case is _scripts/test/Console/Command/Maint/Ticket/InvalidUserCleanup.t_ . In this the following activities take place.

1. A _ZZZZUnitTest*.pm_ file. that sets the temporary article dir, is written to S3
2. The system time is set to '2019-06-15 00:00:00'
3. A _ZZZZUnitTest*.pm_  file, that activates `Ticket::Watcher`, is written to S3

Activity 3. fails with the not very helpful status code 403 and the message "Forbidden". That the activity fails is fine as the S3 interface implements a timestamp sanity check. See "Time stamp requirement" in https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html.

bschmalhofer:
See also #769.

bschmalhofer:
Fixed the first batch of test failures. Let's do a new run before proceeding.
... Looks good so far, down to 14 failing scripts

bschmalhofer:
Same of the tests are failing because of expired test certificates. See #1779. So, I'm ignoring the failures in

- /opt/otobo/scripts/test/SMIME.t
- /opt/otobo/scripts/test/SMIME/EmailHandling.t
- /opt/otobo/scripts/test/Selenium/Agent/Admin/AgentTicketEmailPGP.t
- /opt/otobo/scripts/test/Selenium/Output/Preferences/Customer/SMIME.t

for now,

@svenoe : Do you have insight whether certificates need to be synced via S3?

bschmalhofer:
There are failures with the test test script _scripts/test/Selenium/Agent/Admin/AdminLog.t_ . These failures are kind of random, as errors appear at different lines and sometimes the test scripts succeeds. This is not surprising. _Kernel/Modules/AdminLog.pm_ uses `Kernel::System::Log::GetLog()_ which reads a copy of the log from shared memory. The web containers do not share their shared memory per default. Thus it depends on the container what the content in IPC shared memory is. 

There is a setup where containers share their shared memory, but IMHO this is against the spirit of containerisation,
https://docs.docker.com/engine/reference/run/#ipc-settings---ipc . Thus we simply skip the test when OTOBO_SYNC_WITH_S3 is active.

bschmalhofer:
[...]

PR #2261, opened 2023-04-05
Issue #1777 dynamic field database t

bschmalhofer:


stefanhaerter (approved):
Looks good to me. The assumption of a timing error seems reasonable.

Issue #2220, opened 2023-03-09
Backport features in bin/docker/quick_setup.pl from the rel-10_1 branch


bschmalhofer:
For development it is convenient to automatically set up an example customer and an example calendar when using _quick_setup.pl_. So port back the features added in rel-10_1.  The S3 activation feature will not be ported back.
See also #1929.

bschmalhofer:
Tested by successfully calling the bash alias:
`alias otobo_docker_quick_setup='docker-compose exec web bash -c "date ; hostname ; rm -f Kernel/Config/Files/ZZZAAuto.pm ; bin/docker/quick_setup.pl --db-password otobo_root --http-port 81 --activate-elasticsearch --add-customer-user --add-calendar"'`
PR is merged closing this issue.

bschmalhofer:
Reopening because the test script _scripts/test/Selenium/Agent/AgentAppointmentCalendarOverview.t_ needs to be adapted too. See https://github.com/RotherOSS/otobo/commit/a1e7022da314d3a9fa446cfb2a865f8a9a28815e .

bschmalhofer:
Cherry picked the changes to _scripts/test/Selenium/Agent/AgentAppointmentCalendarOverview.t_ from the rel10_1. The test script no longer fails, even when a calendar already exists. Closing this issue again.

PR #2276, opened 2023-04-14
Issue #2220 backport test script

bschmalhofer:


bschmalhofer:
CodePolicy fails because the checkout fails. Merging anyways, as the PR consists only of cherry-picked commits from rel-10_1.

Edit: it failed because I tried to merge into rel-10_1 instead of rel-10_0.

PR #2290, opened 2023-04-26
Issue #1777 uninitialized value

bschmalhofer:


bschmalhofer:
Merging despite violations of CodePolicy. Looks like the Github action does not use the latest rel-11_0 CodePolicy.

Issue #2303, opened 2023-05-11
Comply with the Perl::Critic policy RegularExpressions::ProhibitCaptureWithoutTest


bschmalhofer:
Checking the OTOBO code with the Perl::Critic policy [RegularExpressions::ProhibitCaptureWithoutTest](https://metacpan.org/module/Perl::Critic::Policy::RegularExpressions::ProhibitCaptureWithoutTest/source) give violations in 30 files.

A couple of reported violations are actually fine,  as there effectively is a check of the previous match that is not recognized by the policy. 

See also:

- https://www.reddit.com/r/perl/comments/13emlxz/failed_matches_dont_reset_the_match_variables/
- https://metacpan.org/dist/Perl-Critic/source/lib/Perl/Critic/Policy/RegularExpressions/ProhibitCaptureWithoutTest.pm

TODO:

- [x] fix the violations where there is no danger of breaking things
- [x] configure **exception_source** of the policy in case there are relevant cases
- [ ] decide whether the policy should be added to CodePolicy

bschmalhofer:
After some easy fixes the number of violations is down to 20 files in rel-11_0.

bschmalhofer:
It would be nice if RegularExpressions::ProhibitCaptureWithoutTest  would not complain about code like this:
```

            for my $SearchStringParamName ( sort @ParamNames ) {
                next SEARCHSTRINGPARAMNAME if $SearchStringParamName !~ m{\ASearchStrings\[(.*)\]\z}sm;
                $SearchStrings{$1} = $ParamObject->GetParam( Param => $SearchStringParamName );
            }

```
On the other hand, there is no complaint about:

```
            for my $SearchStringParamName ( sort @ParamNames ) {
                next SEARCHSTRINGPARAMNAME unless $SearchStringParamName =~ m{\ASearchStrings\[(.*)\]\z}sm;

                $SearchStrings{$1} = $ParamObject->GetParam( Param => $SearchStringParamName );
            }
```

Personally I prefer the version with `unless` anyways, so everything is fine.

bschmalhofer:
Down to 12 files with violations.

bschmalhofer:
Down to 7 files with violations.

bschmalhofer:
[...]

PR #2307, opened 2023-05-13
Issue #2303: eliminate violations of ProhibitCaptureWithoutTest

bschmalhofer:
by using the values returned from using the match operator in list context.


bschmalhofer:
Merging despite CodePolicy violations, as Github does not use the newest version of the CodePolicy.

Issue #2315, opened 2023-05-24
Dynamic Field Object Type and Namespace filtering


stefanhaerter:
Introduce dynamic Field Object Type and Namespace filtering.

stefanhaerter:
AdminDynamicField modules 'forget' the typed field name in error case. This is due to filling the parameter PlainFieldName, which is used in the template, only in Change case. I would suggest to add the following line in the `_ShowScreen` methods:

`$Param{PlainFieldName} //= $Param{Name};`

stefanhaerter:
Closed with merging the separate pull requests mentioned.

PR #2319, opened 2023-05-24
Issue #2315: Enhanced filtering of dynamic fields in admin overview screen.

stefanhaerter:
Related to #2315

### DIFF
--- a/Kernel/System/DynamicField.pm
+++ b/Kernel/System/DynamicField.pm
@@ -734,7 +734,7 @@ sub DynamicFieldList {
     else {
         # create sql query
         my $SQL = 'SELECT id, name, field_order FROM dynamic_field';
-        my @BindValues;
+        my @Bind;
 
         # get database object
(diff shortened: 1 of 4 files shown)


Kernel/System/DynamicField.pm
```diff
@@ -773,20 +773,20 @@ sub DynamicFieldList {
...
             if ( IsStringWithData( $Param{Namespace} ) && $Param{Namespace} ne 'All' ) {
                 push @WhereClauses, 'name LIKE ?';
-                push @BindValues,   "$Param{Namespace}-%";
+                push @Bind,         \"$Param{Namespace}-%";
             }
 
         }
 
-        if ( @WhereClauses && @BindValues ) {
+        if ( @WhereClauses && @Bind ) {
```
bschmalhofer, line 781: The check for @Bind is a bit strange as it isn't used in the if-block.

stefanhaerter: You're right - I changed the code accordingly. Thank you :)

Issue #2295, opened 2023-05-03
Redirect to installer.pl when OTOBO is not installed yet


bschmalhofer:
There had been reports that the installer page could not be opened because of a typo in the URL. This could be avoided by changing the redirect to installer.pl instead of index.pl when OTOBO is not yet installed.

bschmalhofer:
@svenoe and @stefanhaerter 
Could you take a look at #2331 and tell whether this makes sense and is useful?

bschmalhofer:
Discussed this with @svenoe . Let's not do this. Users might be inadvertently directed to installer.pl.  OTOBO admins can be expected to type in installer.pl,

bschmalhofer:
Another case is when index.pl is explicitly opened without having run installer.pl first. See https://github.com/RotherOSS/otobo-docker/issues/171. But the argument from above applies for that case too.

PR #2331, opened 2023-06-04
Issue #2295: redirect to installer.pl

bschmalhofer:
when the specific endpoint is not declared and
ZZZAAuto.pm does not exist

### DIFF
--- a/bin/psgi-bin/otobo.psgi
+++ b/bin/psgi-bin/otobo.psgi
@@ -410,15 +410,29 @@ my $DumpEnvApp = sub {
 
 # Handler for 'otobo', 'otobo/', 'otobo/not_existent', 'otobo/some/thing' and such.
 # Would also work for /dummy if mounted accordingly.
-# Redirect via a relative URL to Frontend::DefaultInterface.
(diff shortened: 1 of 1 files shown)


bin/psgi-bin/otobo.psgi
```diff
@@ -455,7 +469,7 @@ my $RedirectOtoboApp = sub {
     $Res->redirect($Redirect);
 
     # send the PSGI response arrayref
-    return $Res->finalize();
+    return $Res->finalize;
```
stefanhaerter, line 472: Out of curiosity: Is there a technical reason or is this code styling? If it is for styling purposes, should I adapt this syntax?

stefanhaerter (approved):
LGTM - useful improvement in my opinion.

bin/psgi-bin/otobo.psgi
bschmalhofer: There is no special reason, just a matter of taste. I like it because it IMHO emphasizes that this a method call, but that is purely subjective.

bschmalhofer:
#2295 won't be implemented.

Issue #2366, opened 2023-06-23
Document LoopProtection Parameter in Kernel::System::Ticket::FieldRestrictions::GetFieldStates()


stefanhaerter:
I ran across the LoopProtection parameter being not present in the POD. Will make a quick PR to fix this.

bschmalhofer:
PR is applied. Closing this issue.

PR #2367, opened 2023-06-23
Issue #2366: Added LoopProtection to GetFieldStates POD.

stefanhaerter:
Fixing #2366

### DIFF
--- a/Kernel/System/Ticket/FieldRestrictions.pm
+++ b/Kernel/System/Ticket/FieldRestrictions.pm
@@ -66,6 +66,8 @@ sub new {
 
 Returns possible values, selected values, and visibility of fields
 
+    my $LoopProtection = 100;
+
     my %States = $FieldRestrictionsObject->GetFieldStates(
         TicketObject        => $TicketObject,
         DynamicFields       => $DynamicFieldConfigs,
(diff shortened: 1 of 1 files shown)


bschmalhofer (dismissed):
Documenting the current state is good. Approved.

Kernel/System/Ticket/FieldRestrictions.pm
```diff
@@ -80,9 +80,11 @@ Returns possible values, selected values, and visibility of fields
             %GetParam,
             OwnerID     => $GetParam{NewUserID},
         },
+        LoopProtection      => 100,                                 # restricts number of recursive calls; passing 'undef' will lead to a warning
```
bschmalhofer: Can LoopProtection really be 100? My understandig is that a reference to a scalar must be passed.

stefanhaerter: You are right, this has to be a scalar reference. I will adjust the POD accordingly - thank you! =)

bschmalhofer (approved):
Yes, this emphasises that the loop protection variable will be modified within the method.

bschmalhofer:
Moved this to rel-11_0, as adding those minor improvements to maintainance branches is not worth the hassle.

Issue #2375, opened 2023-06-29
new failure in AgentAppointmentCalendarOverview.t  


bschmalhofer:
Test failures are almost gone now, but there is a new failure:
```
/opt/otobo/ scripts/test/Selenium/Agent/AgentAppointmentCalendarOverview.t                                 (Wstat: 257 (exited 1) Tests: 165 Failed: 1)
  Failed test:  165
  Non-zero exit status: 1
```
Looks like some kind of Layout issue, likely not related to the JSON changes. Let's handle this in a new issue.

_Originally posted by @bschmalhofer in https://github.com/RotherOSS/otobo/issues/2371#issuecomment-1613388799_

stefanhaerter:
For me, it looks like the failing test is indicating that the timeline views in CalendarOverviews are broken.

bschmalhofer:
This was needed because of the switch to Cpanel::JSON::XS, see #399 . When an empty string was added to Core.Config, then with the module `JSON`  it was serialized for JS as null. With the module `Cpanel::JSON::XS` the empty string is serialized as the empty string. This broke the `IsRTLLanguage` property used by the JS libraries datepicker and fullcalendar.

bschmalhofer:
The test succeeds now again. A manual tests looked fine too. Closing this issue.

bschmalhofer:
The expression

`( $TextDirection eq 'rtl' ) ? $Kernel::OM->Get('Kernel::System::JSON')->True() : $Kernel::OM->Get('Kernel::System::JSON')->False(),`

is a bit complicated. Let's add a method `Kernel::System::JSON::Bool` that takes a Perl value and turns it into `true` and `false`according to Perl rules. This means that the string `q{0}` will be converted to `false`.

PR #2378, opened 2023-07-05
Issue #2375: Using JSON object True and False methods instead of 1 and

stefanhaerter:
Closing #2375


bschmalhofer:
This was needed because of the switch to Cpanel::JSON::XS. When an empty string was added to Core.Config, then it was previously serialized for JS as null. With Cpanel::JSON::XS the empty string is serialized as the empty string. This broke the IsRTLLanguage property of the datepicker and fullcalendar.

The patch looks fine.

Issue #2363, opened 2023-06-21
Set no value for the boolean HTML attribute checked


bschmalhofer:
I assume this would also apply to checked="checked", right?

_Originally posted by @stefanhaerter in https://github.com/RotherOSS/otobo/issues/2333#issuecomment-1600508988_

bschmalhofer:
Merged the changes into rel-11_0 and issue-#2251-multivalue-multirow.  Close only after the next run of the test suite.

bschmalhofer:
The latest run of the test suite looked fine. Closing now.

PR #2398, opened 2023-07-16
Issue #2363 checked checked

bschmalhofer:


bschmalhofer:
The CodePolicy violations can be ignored, Kernel::System::Web::Request is declared as a soft dependency.

Issue #2475, opened 2023-08-16
Standardize behavior of ValueGet for reference dynamic fields


stefanhaerter:
The value structure used within `sub ValueGet` should be maintained and not be altered by e.g. `sub ValueForLens`.

stefanhaerter:
Closed with merging #2472

bschmalhofer:
Reopen as I see some TODOs in that area.

stefanhaerter:
Which TODOs are left here?

bschmalhofer:
I think that this issue can be considered as being resolved, as there had been independent developments in the mean time. Closing now.

PR #2476, opened 2023-08-16
Issue #2475 lens valueget

stefanhaerter:
Closing #2475


stefanhaerter:

Issue #2333, opened 2023-06-07
Set no value for the boolean HTML attribute 'readonly'


bschmalhofer:
Boolean attributes are true if present, they require no value. But setting the attribute name as the value is fine. See https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes .
My understanding is that setting no value is the better practice. In the OTOBO I have seen both variants. Thus I propose to standardize on setting no value.

The same applies to CSS. The following CSS selectors can be used: `input:not([readonly]):hover` and `input:[readonly]:hover` .

stefanhaerter:
I assume this would also apply to checked="checked", right?

bschmalhofer:
Yes, this can be done in a dedicated issue.

bschmalhofer:
The unit tests showed no new failures. PR is merged. Closing this issue.

stefanhaerter:
There are some leftovers in the css where elements are selected based upon `[readonly="readonly"]`, which (most likely, double-checking this) fails. I'll provide a PR for this soon.

bschmalhofer:
It wasn't really clear to me whether [readonly="readonly"] still matches. My impression was that it was fine, and my hope was that this isn't browser dependent.

stefanhaerter:
To be honest, I didn't think of browser dependance - I encountered the problem when setting a dynamic field readonly via ticket mask definition, which did work as expected in agent frontends, but (at least the css) not in customer frontend despite the field having the correct class and getting the readonly attribute. Maybe the problem is that my installed version of firefox appends to the readonly and checked attributes: readonly` in tt file becomes `readonly=""` in browser inspector. I assumed that this was the solution, that `[readonly="readonly"]` does not match `readonly=""` - also, there are cases of `[readonly=readonly]`, I currently do not know if the double quotes make a difference, but I thought the easiest solution would be to unify this also in CSS, but now I am not 100% sure anymore - what do you think?

bschmalhofer:
Unifying it in the CSS should be done anyways, as it reduces confusion. Looking at the Spec I found that only the existence of the attribute matters, so even readonly="0" should set the attribute.

stefanhaerter:
Thank you for checking =) I'll do the changes then.

stefanhaerter:
@bschmalhofer Do you think the issue can be closed again?

bschmalhofer:
Yes, I'm closing the issue again.
[...]

PR #2482, opened 2023-08-19
Issue #2333 unify css

stefanhaerter:
Closing #2333 again


bschmalhofer (approved):
Looks good.
I alos checked where `$ReadonlyInternalField`is used in the templates and found no case where that boolean attribute would be attache to another attribute and break the HTML.

Issue #81, opened 2020-06-08
installer.pl should warn about unsupported database version


bschmalhofer:
A version check of the database would be nice to have.

stefanhaerter:
Could the get_info method be of use here? https://metacpan.org/pod/DBI#get_info
As far as I understand the code, the version check could be placed in Kernel/Modules/Installer.pm, function CheckDBRequirements. Is this correct?

bschmalhofer:
Yes, using a database driver independent method sounds like a good idea. In case that doesn't suffice there are also driver specific methods and attributes, like https://metacpan.org/pod/DBD::mysql#Private-MetaData-Methods.

CheckDBRequirements() looks like a good place where this check can be added.

stefanhaerter:
In a post from perlmonks it was stated that the get_info method was (only?) for OCDB. The post origins from 2010, so I hope that the method was adapted to other database drivers in the meantime. I will research for further information on this possibility.

bschmalhofer:
This is an old issue but still desirable. I propose to finally implement it for OTOBO 11.0.   Using the approach taken by @daxim in https://github.com/RotherOSS/otobo/commit/e060b61f450d0cd0a574dae32bd1dfaf4f217352 but using only [version.pm](https://metacpan.org/dist/version/view/lib/version.pod)

bschmalhofer:
The change is merged. Let's keep this issue open, in case any problems crop up.

jduke-halls:
Thinking as a DBA, in MySQL and MariaDB there's the variable @@version.  Would this be useful for this instance?  When running on my OTOBO server:

`mysql> select @@version;`
`+-------------------------+`
`| @@version               |`
`+-------------------------+`
`| 8.0.31-0ubuntu0.22.04.1 |`
`+-------------------------+`

bschmalhofer:
Yes, this should be the same string that is returned by [DBI::get_info](https://metacpan.org/pod/DBI#get_info). Another way of getting the version string is `SELECT VERSION()`. This way is effectively used by the SupportDataCollector.

bschmalhofer:
[...]

PR #2505, opened 2023-09-15
Issue #81: strip '5.5.5-' from the reported MariaDB version

bschmalhofer:
This prefix was added only to allow replication with MySQL.


bschmalhofer:
Ignoring the violation of CodePolicy, as the CodePolicy on Github is not up to date.

Issue #2507, opened 2023-09-15
Suggestion: Change label handling for dynamic fields to bypass indexing problem


stefanhaerter:
I would want to propose to handle the labels for dynamic fields in the same way as the field html itself. This surpasses the problem of input error handling in js, which occurs because the label `id` and `for` attributes do not match the field id, except for the very first field rendered (to the best of my knowledge).

stefanhaerter:
This issue has been stale for quite some time now. If the topic surfaces again, please feel free to reopen, but closing it in the meantime. Also, the code base has changed a lot since I started working on the PR.

PR #2508, opened 2023-09-15
Issue #2507 df label proposal

stefanhaerter:
Proposal for #2507 

Important notice: The process modules will only work in combination with PR #2506 because the previous hash keys and structure there is not in line with the current state of the dynamic field development.


stefanhaerter:
See comment in issue, closing the PR as well.

PR #2522, opened 2023-09-24
Invalid agents can be included in ticket search via sysconfig option.

stefanhaerter:
For rel-10_1, see package proposal [here](https://git.gedankenblogsammlung.de/demiguise/ticketsearch-includeinvalidagents).


stefanhaerter:
It needs to be discussed if we actually want to include this possibility. Leaving the PR as pending in the meantime.

stefanhaerter:
Closing the PR as the topic did not pop up again.

Issue #2251, opened 2023-04-03
Multi value and multi row dynamic fields.


svenoe:
Introduce multi value and multi row dynamic fields.

stefanhaerter:
Pushed the current status.

stefanhaerter:
- [x] Add frontend_mask_definition to database schema
- [x] Create migration script for rel-11_0

bschmalhofer:
TODO:

- [ ] satisfy the unit tests, currently there are many failures
- [ ] let Backend.pm convert simple scalar values into an array ref, the drivers get always array refs as the value

bschmalhofer:
For convenience it would be nice that when adding a multifield the cursor would jump into the new input field.

bschmalhofer:
A small bug: all the hidden or shown labels of multivalue inputs have the same 'id' and 'for' attribute.

bschmalhofer:
Trying to figure out the failing test case in _scripts/test/DynamicField/DynamicFieldDBViaWebService.t _ . This is a bit confusing. There is the succeeding test "Set Database DynamicField value to incorrect user" where two values are sent to the webservice.

```
DebugLog debug:
  Summary: Using invoker 'TicketUpdate'
  Data   : No data provided.
DebugLog debug:
  Summary: Outgoing data before mapping
  Data   : $VAR1 = {
  'DynamicField' => [
    {
      'Name' => 'Databasetest4183004540100000',
      'Value' => [
        'test4183004540100001',
        'test123'
      ]
    }
  ],
  'Password' => 'test4183004540100001',
  'TicketID' => 56,
  'UserLogin' => 'test4183004540100001'
};
```
But then there is the failing test case "Set Database DynamicField value to multiple users - request result is correct" where also two values are sent over, but this time the values are wrapped into a hashref. The reason for the difference is not obvious.

```
DebugLog debug:
  Summary: Using invoker 'TicketUpdate'
  Data   : No data provided.
DebugLog debug:
  Summary: Outgoing data before mapping
  Data   : $VAR1 = {
  'DynamicField' => [
    {
      'Name' => 'Databasetest4183004540100000',
      'Value' => [
        { ValueText => 'test4183004540100001' },
        { ValueText => 'test123' },
      ]
    }
  ],
  'Password' => 'test4183004540100001',
  'TicketID' => 56,
  'UserLogin' => 'test4183004540100001'
};

```

Edited: fixed the Value attribute of the second data structure,

[...]

PR #2524, opened 2023-09-25
Issue #2251 readonly enhancemenets

stefanhaerter:
Referencing #2251


stefanhaerter:
Closing this as the proposed distinction between readonly and disabled can't be used due to technical reasons.

PR #2527, opened 2023-09-26
Issue #1531: Consider missing or empty parents as error when treeview is enabled in dynamic select fields.

stefanhaerter:
Closing #1531


stefanhaerter:
Re-doing this for 10.0 because the issue is for milestone 10.0.

Issue #991, opened 2021-05-06
Bug In SysConfig - Keys with ### do not work in the frontend (used in DynamicFieldScreens)


StefanRother-OTOBO:
Hi,

If I go to Admin -> System Configuration and click on Edit for the option "DynamicFieldScreens###Framework“ for example and then click on Save without making any changes, changes are still written. DynamicFieldSreens don't work anymore.

Thanks!

Stefan

svenoe:
Actually there are changes made - the keys are set to broken values. You can reset the setting to fix it, and you could put in every field by hand and it would work, but that is nonsense of course. I'll look where it comes from.

Edit: Slight correction - it wouldn't work. A part of the string is duplicated during pressing OK, basically...

svenoe:
The error is produced in the js-part, Core.SysConfig.js line 429 (in Core.SysConfig.Update()):
`FullName = FullName.substr(0, FullName.lastIndexOf("###"));`
The "###" at the moment cannot be part of any hash key in the SysConfig. I assume it's the same in OTRS. A quick solution would be to change this to lastIndexOf("_Hash###"), but it's still not really clean. I want to at least consider some other possibilities, have to think about it, though...

stefanhaerter:
Can it be assumed that the wanted FullName does not contain more than two occurrences of ###, like in `DynamicFieldScreens###Framework_Hash###Ticket::Frontend::AgentTicketClose###DynamicField`? Then maybe it could a possibility to do something like this:

`FullName = FullName.substr(0, FullName.indexOf(FullName.firstIndexOf('###') + 1, '###'));`

This would find the second occurrence of ###.

stefanhaerter:
Another suggestion I thought of:

Replace

`FullName = FullName.substr(0, FullName.lastIndexOf('###'));`

with

`FullName = FullName.substr(0, FullName.indexOf('###' + Key));`

However, I currently do not fully understand the purpose of removing the key from the full name (which at the start includes the key, at least in the examples I tested), just to add it again. 

Edit: This is about the following code section:

https://github.com/RotherOSS/otobo/blob/9a6e290ad2a3137576d9940b16026358d8fd9292/var/httpd/htdocs/js/Core.SystemConfiguration.js#L425-L431

stefanhaerter:
Closed with merging #2533

PR #2528, opened 2023-09-26
Issue #991: Check and trim sysconfig key if it contains hashes.

stefanhaerter:
Closing #991


stefanhaerter:
Re-doing the PR for rel-10_1.

PR #2554, opened 2023-10-05
Issue #1946: Prevented using unitialized password value for checking if to resetting preferences.

stefanhaerter:
Re-done PR.

Closing #1946

### DIFF
--- a/Kernel/Modules/AdminCustomerUser.pm
+++ b/Kernel/Modules/AdminCustomerUser.pm
@@ -371,7 +371,9 @@ sub Run {
                 );
             }
 
-            if ( $GetParam{UserPassword} && $CurrentUserData{UserPassword} ne $GetParam{UserPassword} ) {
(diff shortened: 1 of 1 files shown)


Kernel/Modules/AdminCustomerUser.pm
bschmalhofer, line 1: The change actually changes the logic in the case where `$CurrentUserData{UserPassword}` contains a false value, e.g. the string '0'. I suggest to only eliminate the warning by writing something like;
 (  $CurrentUserData{UserPassword} // '') ne $GetParam{UserPassword}

stefanhaerter: You are right - I will rewrite this. To be honest, I was a bit too quick as I tried to clean up my backlog of branches and stashes ;) Thank you for the helpful comment!

stefanhaerter:
Closing in favor of #2556

Issue #2298, opened 2023-05-05
New dynamic field types: Reference and Lens


bschmalhofer:
Two new field types are deemed to be useful for OTOBO 11.0.x:

**Reference**
A dynamic field that references another entity, like Ticket, FAQ, or Configitem. Only entities supporting dynamic fields can be linked. The config of an instance of ReferenceDF holds information about the type of the linked objects and optionally further restrictions, like the class of a config item. This is a bit more restrictive than linking via the link table. The reference entity can't be changed via the dynamic fields.

**Lens**
The name refers to https://en.wikipedia.org/wiki/Bidirectional_transformation. It allows to read and write a dynamic field of an entity that is available via the Reference DF. Not sure whether that really is bijective. The design explicitly allows for direct manipulation of the internals of another object. So care must be taken when defining a dynamic field of the type Lens. The config of Lens DFs hold dynamic field id of a Reference DF and the dynamic field id of a DF that is supported by the Reference DF.

bschmalhofer:
This feature is based on the enhancements done for #2251.

bschmalhofer:
I noticed that dynamic fields types are declared in different XML files. This makes sense for specialized dynamic fields types, like the types used by process management. But it makes no sense that generally usable dynamic fields like **Text** or **Multiselect** are declared in _Ticket.xml_.

StefanAbel-OTOBO:
Please make sure that the objects are actually linked to each other (so that you can scroll down to the link objects and reach them).
(I'm just trying to make sure that I have the same understanding of "linked".)

svenoe:
> Please make sure that the objects are actually linked to each other (so that you can scroll down to the link objects and reach them). (I'm just trying to make sure that I have the same understanding of "linked".)

No, they won't be linked via standard links. (We can think about an extension which does this later, but not currently.)

svenoe:
> I noticed that dynamic fields types are declared in different XML files. This makes sense for specialized dynamic fields types, like the types used by process management. But it makes no sense that generally usable dynamic fields like **Text** or **Multiselect** are declared in _Ticket.xml_.

As written before, the dynamic fields in Ticket.xml are the OTRS standard dynamic fields. The config of special dynamic fields like Webservices etc., which were added with OTOBO were written to a separate file. This is not consistent and can be improved, but it has no high priority either.

bschmalhofer:
Test suite looks mostly fine:

[...]

PR #2577, opened 2023-10-18
Issue #2298: Disable referenced object type dropdown in admin dynamic field reference.

stefanhaerter:
Referencing #2298


stefanhaerter:
Closing because the referenced object type is needed in web request params, which would be prevented by disabling the selection.

Issue #2371, opened 2023-06-24
New test failure in AdminSystemConfigurationExampleSessionID.t 


bschmalhofer:
Since 2023-05-18, that is the first test run with Cpanel::JSON::XS, the test _scripts/test/Selenium/Agent/Admin/AdminSystemConfigurationExampleSessionID.t_ reports an error. Here is the relevant snippet.

 
```
103         # Verify the deploy link click.
104         $Selenium->WaitFor(
105             ElementExists => '//a[contains(@href,"Subaction=Deployment")]',
106         );
107         $Selenium->find_element('//a[contains(@href,"Subaction=Deployment")]')->VerifiedClick();
108 $DB::single = 1;
109         $Self->Is(
110             $Selenium->execute_script("return \$('#DeploymentStart').length > 0"),
111             "1",
112             "The deployment link not redirecting to login.",
113         );

```
It is not obvious what the correct behavior is. The test script does:

-  turns off the session cookie
-  log in with credentials
- change a sample SysConfig setting (all in JS presumably)
- checks whether the alert for the deployment shows up
- clicks on the link in the alert
- expects that there is no redirect to the login page

The last assertion fails, there is a redirect to the login page. This makes sense, as the session cookie is turned off.

TODO:

- [x] investigate why it worked before
- [x] fix the code or adapt the test suite

bschmalhofer:
Boiled it down to a single commit. The error occurs in the commit https://github.com/RotherOSS/otobo/pull/2338/commits/a5e323305446375d5dd2d43c07d1161c7a231bc5  but not in a5e323305446375d5dd2d4^. The difference is not really apparent. In both cases the test script should click on the notification for dirty SysConfig, which is the link   https://localhost:1484/otobo/index.pl?Action=AdminSystemConfigurationDeployment;Subaction=Deployment . But only a5e323305446375d5dd2d4^ then redirects to the login page.

On possibility is that in the working case there is some kind of JavaScript magic that makes no request to the server. The next step could be a comparison of the HTML that shows the notification.

See #399.

bschmalhofer:
Might be related to https://github.com/OTRS/otrs/commit/bd974e5c8e2c4f7f6a3a4ddefa32a3cbbee9c1f1#diff-8e6b4bce22cf7471e867d1b768cb1d147df1263aec09298ace8ae4cdccec2298.  The session id should be added to the deployment link.

bschmalhofer:
[...]

PR #2593, opened 2023-10-24
Issue #2371: revert changes in the Get() method

bschmalhofer:
Returning the default value is now triggered again by the truthiness of the stored value. Note that the string containing the digit 0 is true.

### DIFF
--- a/var/httpd/htdocs/js/Core.Config.js
+++ b/var/httpd/htdocs/js/Core.Config.js
@@ -85,12 +85,20 @@ Core.Config = (function (TargetNS) {
      * @memberof Core.Config
      * @function
      * @returns {Object} The value of the option. Can be every kind of javascript variable type. Returns undefined if setting could not be found.
-     * @param {String} Key            - The name of the config option. May also be a combined option like Richtext.Width.
-     * @param {Object} [DefaultValue] - If nothing is saved in the config, return this default value. The DefaultValue may be omitted.
+     * @param {String} Key            - The name of the config option. May also
+     *                                  be a combined option like Richtext.Width.
+     * @param {Object} [DefaultValue] - If nothing or a false value is saved in the config
+     *                                  then this default value is returned. The DefaultValue may be omitted.
      *                                  In that case undefined is returned.
-     *                                  The saved values 0, '', false are supported.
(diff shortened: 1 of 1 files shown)


bschmalhofer:
That the change has to be reversed is clear. However there are five new test failures in the test suite, including the original errorAdminSystemConfigurationExampleSessionID.t. I merged anyways and accept that there are some intermediary failures.

Here is the analysis part. It was already stated above that the problem can be highlighted by dumping a numeric Sysconfig setting one time with `JSON` and the other time `Cpanel::JSON::XS`. 

```
otobo@ad98e2ba4b0f:~$ cp JSON_old_but_working.pm Kernel/System/JSON.pm 
otobo@ad98eis2ba4b0f:~$ bin/otobo.Console.pl Maint::Config::Dump SessionUseCookie
"0"
otobo@ad98e2ba4b0f:~$ cp JSON_new_but_broken.pm Kernel/System/JSON.pm 
otobo@ad98e2ba4b0f:~$ bin/otobo.Console.pl Maint::Config::Dump SessionUseCookie
0
otobo@ad98e2ba4b0f:~$ 
```
Further debugging showed that the versions of the JSON libraries do not matter.

The explanation why `SessionUseCookie`  is considered one time as a string and the other time as number is not obvious. But the relevant hint is right in the documentation of `Cpanel::JSON::XS`: https://metacpan.org/pod/Cpanel::JSON::XS#simple-scalars .

> Non-deterministic behavior is following: scalars that have last been used in a string context before encoding as JSON strings, and anything else as number value:

For SessionUseCookie this checks out. The number is created by running ZZZAAuto,pm and never meddled with, e.g. printed, before converted to JSON.

 `JSON::XS` follows more or less the same strategy but apparently there are differences in details.

The take home message is that the decision why a (numeric) value is represented as JSON as a string might be deterministic, but not transparent. So the next best idea is to force transparency. This can be done by setting `$json->type_all_string(1)` for the values that are passed into the JavaScript World. This is a simple rule, that forces stringiness on the numbers. It seems to be in line with the current implementation:

```
otobo$ grep -r Core.Config.Get  var/httpd/htdocs/js/ | grep parseInt | wc -l
27
```

PR #2600, opened 2023-10-26
Issue #2295: revert to JSON::XS

bschmalhofer:
It was attempted to reproduce the JSON::XS behavior with Cpanel::JSON::XS. The option type_all_string looked promising for that. But using that option would require a version of Cpanel::JSON::XS that is not provided per default by the current LTS version.

This also means that boolean can now be passed as real booleans to JavaScript.


bschmalhofer:
Ignoring the complaint about Types::Serialiser::true by CodePolicy.

PR #2603, opened 2023-10-26
Issue #2298: Using correct variable type to store function result.

stefanhaerter:


stefanhaerter:
Closing because this seems to be the wrong place for the fix. Further investigating.

PR #2645, opened 2023-11-06
Update TicketStateSet.pm

StefanAbel-OTOBO:
Added PendingTime and PendingTimeDiff as default options in TicketStateSet Transition Action

### DIFF
--- a/Kernel/System/ProcessManagement/TransitionAction/TicketStateSet.pm
+++ b/Kernel/System/ProcessManagement/TransitionAction/TicketStateSet.pm
@@ -83,6 +83,16 @@ sub Params {
             Value    => '1 (can overwrite the logged in user)',
             Optional => 1,
         },
+        {
+            Key      => 'PendingTime',
+            Value    => '2023-11-06 18:02:00',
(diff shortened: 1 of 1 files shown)


Kernel/System/ProcessManagement/TransitionAction/TicketStateSet.pm
```diff
@@ -83,6 +83,16 @@ sub Params {
             Value    => '1 (can overwrite the logged in user)',
             Optional => 1,
         },
+        {
+            Key      => 'PendingTime',
```
svenoe, line 87: PendingTimeDiff is a valid parameter, but PendingTime doesn't seem to be one. Could you please quickly confirm, that it works in the standard?

stefanhaerter: PendingTime and PendingTimeDiff seem to be used and documented for TicketCreate, so I would consider both feasible. But I will have them tested.

stefanhaerter:
Apparently this change has already been added in #3136 - closing the PR.

Issue #2630, opened 2023-10-31
Miscellaneous changes related to dynamic field rendering in frontends


stefanhaerter:

stefanhaerter:
Closed with merging #2632

PR #2663, opened 2023-11-10
Issue #2630 df valuerender

stefanhaerter:
Referencing #2630 

This PR holds multiple small but relevant changes. I will try to give a brief summary:

- As I encountered the problem of using undefined values in DisplayValueRender and ReadableValueRender multiple times, I took the freedom to align the handling at every relevant occurrence with the drivers where it works
- CustomerTicketProcess now passes its process suffix to ValueSet - this is because for Lens fields, EditFieldValueGet is called during the procedure and this fails to retrieve the value from the web request without the process suffix
- AgentTicketProcess assumed in its AJAXUpdate handling that the first input of multivalue fields still has no index (DynamicField_Name vs. DynamicField_Name_0) as it has been before - I adapted the behavior to the current naming
- When adding a new reference dynamic field, EditFieldMode is undef which caused a warning in the log - for this case, I changed the value to an empty string
- Adjusted some POD and comments

### DIFF
--- a/Kernel/Modules/AdminDynamicFieldReference.pm
+++ b/Kernel/Modules/AdminDynamicFieldReference.pm
@@ -780,6 +780,7 @@ sub _ShowScreen {
     );
 
     # compute value for editfieldmode to maintain frontend selection
+    $Param{EditFieldMode} //= '';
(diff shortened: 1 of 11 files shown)


stefanhaerter:
Closing in favor of #2667

Issue #1197, opened 2021-08-26
Support or switch to JavaScript::Minifier::XS


bschmalhofer:
To me it looks like this was related to CKEditor 4.15.1 and does not occur with 4.16.0. We should consider switching to the XS versions anyways, I guess. I will postpone it to 10.0.9 though.

_Originally posted by @svenoe in https://github.com/RotherOSS/otobo/issues/771#issuecomment-775852304_

bschmalhofer:
I don't see the advantages from having the test and support both JavaScript::Minifier and JavaScript::Minifier::XS. I propose to switch to the XS in 10.1. Always happy to clean out _Kernel/cpan-lib_.

svenoe:
Please have a look whether the XS library is available in the Debian/Ubuntu/etc. repos...

bschmalhofer:
I checked whether the JavaScript::Minifier::XS is available in the relevant distributions. Yes, it is.

bschmalhofer:
In the same vein, switch to CSS::Minifier::XS.

bschmalhofer:
Tests look fine after adapting the expected output. Looks like JavaScript::Minifier::XS keeps the Copyright statements, which makes sense to me.
Closing this issue.

PR #2683, opened 2023-11-16
Issue #1197: require JavaScript::Minifier::XS

bschmalhofer:
and remove JavaScript::Minifier from Kernel/cpan-lib


bschmalhofer:
Forgot to actually switch to JavaScript::Minifier::XS.

PR #2720, opened 2023-11-29
agent ui design

chzauleck:


var/httpd/htdocs/js/Core.UI.InputFields.js
```diff
@@ -247,6 +247,9 @@ Core.UI.InputFields = (function (TargetNS) {
             $('.DynamicFieldReference', $Context).each(function () {
                 Core.Agent.DynamicFieldReferenceSearch.InitElement($(this), AutoCompleteActive);
             });
+
+            //Add an extra class to Fields containing a Dynamic Field Set
+            $('.DynamicFieldSet').parent().addClass('DFSetOuterField')
```
svenoe, line 252: I would add a ", $Context" in the selector (see line 247), I think. Doesn't change anything atm, but just for accuracy.

var/httpd/htdocs/skins/Agent/default/css/Core.Form.css
```diff
@@ -704,26 +704,67 @@ fieldset.TableLike > .Row.MultiColumn > .FieldCell > .Field {
...
-    min-height: 20px;
+/* Dynamic Field Set */
+
+.DFSetOuterField {
+    max-width: 622px; 
 }
 
-fieldset.TableLike fieldset.DynamicFieldSet > .Row.MultiValue > .FieldCell > .Field {
-    display: inline-block;
+.DynamicFieldSet {
```
svenoe, line 713: Be careful, I'm not sure, whether this affects other places like TicketZoom, ConfigItemZoom etc. I would keep the fieldset.TableLike structure here, and maybe add a .Row somewhere outside, too.

Issue #2711, opened 2023-11-27
Integrate ImportExport into OTOBO core


bschmalhofer:
Importing and exporting data is a basic feature. Therefore let's integrate the extension package ImportExport into the core.

TODO:

- [x] check CodePolicy in rel-10_1
- [x] Update the translations in _Kernel/Language/_ based on the changes in _i18n_
- [x] merge the updated rel-10_1 into master
- [x] merge the ImportExport files into OTOBO-core, skipping the translations and the package setup
- [x] keep _Kernel/Config/Files/XML/ImportExport.xml_ as a separate file
- [x] no need to integrate the files in _doc_, they just contain the database schema (still with OTRS 6 labels) 
- [x] Add the tables to _scripts/database/otobo-schema.xml_
- [x] integrate the package setup (actually no setup code is run, thus nothing to do)
- [x] Update the migration package
- [x] merge the translations ( concatenate to the existing files ??? )
- [ ] Update the dependency list of the ITSM bundle package. _Will be done later, as there is no rel-11_0 branch for that module yet_.
- [x] Move the ITSM specific Console Commands to ITSMConfigurationManagement, _Kernel/System/Console/Command/Admin/ITSM/_
- [x] move the tests and the test sample file for the console command to ITSMConfigurationManagement
- [ ] Decide whether ImportExportCustomerCompany and the other ImportExport modules should be integrated as well _Will be decided later_
- [x] Update the installation manual for OTOBO 11.  See https://github.com/RotherOSS/doc-otobo-installation/issues/123

bschmalhofer:
Using msgadd from the gettext project for merging translations. Here is the script, just for future reference:
[integrate_importexport_po_files.sh.txt](https://github.com/RotherOSS/otobo/files/13488217/integrate_importexport_po_files.sh.txt)

The script does a slight reformatting when a translation contains newlines. This reformatting has been ignored when committing the changed .po files.

EDIT: this apparently belongs to some other issue

bschmalhofer:
@svenoe  pointed out that ImportExport should not show up in the list of installed packages when upgrading from 10.1. This functionality has been added to _scripts/DBUpdate-to-11.0.pl_,

Note that this does not work well when ImportExport has been installed in a OTOBO 11 system and then _scripts/DBUpdate-to-11.0.pl_ is called. This is because the .save files are not copied to the overridden files. But this situation can only occur during development, as no version of ImportExport will be released for OTOBO 11.

[...]

PR #2723, opened 2023-11-29
Issue #2711 importexport

bschmalhofer:


bschmalhofer:
The CodePolicy violations were about the *.po files. They should be checked with the command `msgfmt`, but that command is not installed on the GitHub runners. So, let's ignore these violations.

Issue #2731, opened 2023-11-30
Clean up remnants of MSWiin32 support


bschmalhofer:
I was running `grep -r -i Win32` and I found only two remnants of MSWin32 support. Let's clean them up nevertheless.

> Kernel/Modules/AdminDynamicFieldDB.pm:    if ( $^O ne 'MSWin32' ) {
> scripts/test/Package.t:  <OS>MSWin32</OS>
>

bschmalhofer:
Also 

- _Kernel::System::Crypt::PGP.pm:  ` if ( $^O =~ m/Win/i ) {`

stefanhaerter:
Recently, I ran about microsoft xml references (if I interpreted the code correctly) in these files:

- `Kernel/Output/HTML/Templates/Standard/AgentTicketSearchOpenSearchDescriptionFulltext.tt`
- `Kernel/Output/HTML/Templates/Standard/AgentTicketSearchOpenSearchDescriptionTicketNumber.tt`
- `Kernel/Output/HTML/Templates/Standard/CustomerTicketSearchOpenSearchDescription.tt`

Do these also belong to Win32 or something similar?

bschmalhofer:
PR is applied, closing this issue.

PR #2733, opened 2023-11-30
Issue #2731 ms win32

bschmalhofer:


stefanhaerter (approved):
I am not familiar with these parts of the code, but as far as I understand the changes, they look good to me :)

Issue #2313, opened 2023-05-23
Dynamic Field Sets


svenoe:
Introduce dynamic field sets.

bschmalhofer:
Also: the script _scripts/DBUpdate-to-11.0.pl_ needs to handle the upgrade of dynamic_field_value.id from INTEGER to BIGINT. In the case of MariaDB the real data types are int(11) and bigint(20).

PR #2806, opened 2024-01-09
Issue #2313 grids in dynamicfieldsets

chzauleck:


Kernel/Modules/AdminDynamicFieldSet.pm
```diff
@@ -137,6 +137,9 @@ sub _AddAction {
         }
     }
 
+    my $DynamicFieldObject        = $Kernel::OM->Get('Kernel::System::DynamicField');
+    my $DynamicFieldBackendObject = $Kernel::OM->Get('Kernel::System::DynamicField::Backend');
+
```
svenoe: Why was this shifted further up?

Kernel/Modules/AdminDynamicFieldSet.pm
```diff
@@ -148,48 +151,97 @@ sub _AddAction {
...
+                # DF definition for this name exists in DB
                 my $DynamicField = $DynamicFieldObject->DynamicFieldGet(
-                    Name => $Line->{DF},
+                    Name => $DFElement,
                 );
-                if ( !IsHashRefWithData($DynamicField) ) {
+                if ( !$DynamicField ) {
                     $Errors{IncludeServerError}        = 'ServerError';
-                    $Errors{IncludeServerErrorMessage} = Translatable( 'No valid dynamic field "' . $Line->{DF} . '".' );
+                    $Errors{IncludeServerErrorMessage} = Translatable( 'No dynamic field "' . $DFElement . '".' );
```
svenoe: Field could exist but be set to invalid, I would leave the "valid".

Kernel/Modules/AdminDynamicFieldSet.pm
```diff
@@ -148,48 +151,97 @@ sub _AddAction {
...
+                            if ( !$CheckDFElement->($DFEntry->{DF}) ) {
+                                last LINE;
+                            }
+                        }
+                        
+                    }
+
+                    push @Include, { Grid => $Line->{Grid} };
+
+                } elsif ( $Line->{DF} ) {
```
svenoe: Not critical, but also not 100% consistent. For now, I would force that each $Line contains either a grid or a DF, thus change `elsif ( $Line->{DF} )` to a simple `else`. Also it would be nice to somewhere (probably in `&$CheckDFElement`) catch the possibility that e.g. `$DFEntry->{DF}` is `undef`.

Kernel/Modules/AdminDynamicFieldSet.pm
```diff
@@ -430,38 +482,90 @@ sub _ChangeAction {
         );
 
         if ( IsArrayRefWithData($IncludeFrontend) ) {
```
svenoe, line 523: As the whole check is done at two places and it grew a bit, one could think about moving the complete content of this block into a sub. The only "downside" is, that `%Errors` should not be altered in there directly, in this case and things would have to be slightly rewritten.
```
if ( IsArrayRefWithData($IncludeFrontend) ) {
    my %YAMLErrors = $Self->_CheckInclude( $IncludeFrontend );
    ...
}
```

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -81,28 +81,98 @@ sub new {
     return $Self;
 }
 
+# Helper Function for getting the Dynamic Fields from an Include, i.e.
+# $DynamicFields = $GetIncludedDynamicFields->($Param{DynamicFieldConfig}{Config}{Include});
+# This subroutine takes two arguments:
+# $Include: a list of hash references containing information about the items to include
+# $DynamicFieldObject: an object used to retrieve dynamic field information
+# and returns either the DynamicFields or undef in case of an error.
+my $GetIncludedDynamicFields = sub {
```
svenoe: While this is possible, I have not seen anonymous subs used outside of subs anywhere else in OTOBO, and I'm not sure why you would want that here. If there is no real reason, please rewrite it to:
```
sub _GetIncludedDynamicFields {
...
}
```
Also take a look at other places on how to write pod, e.g. https://github.com/RotherOSS/otobo/blob/40a68a17babe7254279c3f27302f037ad8129269/Kernel/System/Ticket.pm#L8003-L8015 or https://github.com/RotherOSS/otobo/blob/40a68a17babe7254279c3f27302f037ad8129269/Kernel/System/Console/BaseCommand.pm#L881-L891

(Actually we could decide to consistently use one of the two versions, but this is not and has not been in a consistent state.)

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -81,28 +81,98 @@ sub new {
...
+        my ($DFEntry, $DynamicFieldObject) = @_;
+        
+        my $DynamicField = $DFEntry->{Definition} // $DynamicFieldObject->DynamicFieldGet(
+            Name => $DFEntry->{DF},
+        );
+
+        return $DynamicField;
+    };
+
+    my $Error = undef;
```
svenoe: As Bernhard noted, this is equivalent to just `my $Error;`.

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -81,28 +81,98 @@ sub new {
...
 
+# Helper Function for getting the Dynamic Fields from an Include, i.e.
+# $DynamicFields = $GetIncludedDynamicFields->($Param{DynamicFieldConfig}{Config}{Include});
+# This subroutine takes two arguments:
+# $Include: a list of hash references containing information about the items to include
+# $DynamicFieldObject: an object used to retrieve dynamic field information
+# and returns either the DynamicFields or undef in case of an error.
+my $GetIncludedDynamicFields = sub {
+
+    my ($Include, $DynamicFieldObject, $DynamicFieldName) = @_;
```
svenoe: In most cases this should be passed as, then,
```
my $DynamicFields = _GetIncludedDynamicFields(
    Include => $Include,
    DynamicFieldName => $DynamicFieldName,
    DynamicFieldObject => $DynamicFieldObject,
);
```
and **could** be checked similar to e.g. https://github.com/RotherOSS/otobo/blob/40a68a17babe7254279c3f27302f037ad8129269/Kernel/System/Ticket.pm#L420-L432
(in this case probably not necessary, just as comment to note; in general for internal functions this is less important and rarely done)

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -81,28 +81,98 @@ sub new {
...
+    my $Error = undef;
+
+    ITEM:
+    for my $IncludeItem (@{ $Include }) {
+
+        if ($IncludeItem->{Grid}) {
+
+            for my $Row (@{ $IncludeItem->{Grid}{Rows} }) {
+
+                for my $DFEntry (@$Row) {
```
svenoe: This is also rarely to never used. Please rather use either "the old way" `@{ $Row }`, or alternatively `$Row->@*` (which does the same).

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -81,28 +81,98 @@ sub new {
...
+    my ($Include, $DynamicFieldObject, $DynamicFieldName) = @_;
+    my @DynamicFields;
+
+    # This subroutine takes a DFEntry and the DynamicFieldObject as arguments
+    # It retrieves the dynamic field definition for the given DFEntry
+    # If the definition is not available, it retrieves it from the DynamicFieldObject
+    # Returns the dynamic field definition
+    my $GetDynamicField = sub {
+
+        my ($DFEntry, $DynamicFieldObject) = @_;
```
svenoe: As you are defining the sub within the scope of `GetIncludedDynamicFields` and `$DynamicFieldObject` is already initialized in line 92, you do not need to pass it to `$GetDynamicField`.

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -81,28 +81,98 @@ sub new {
...
+        } else {
+            $Error = 1;
+            last ITEM;
+        }
+    }
+
+    if ($Error) {
+        $Kernel::OM->Get('Kernel::System::Log')->Log(
+            Priority => 'error',
+            Message  => "$DynamicFieldName configured erroneously. 'DF: DynamicFieldName' missing or wrong.",
```
svenoe: You could set `$Error` to either "No DynamicField '$IncludeItem->{DF}'.", or "Need either Grid or DF in each Row.", or similar, based on the context, and then set the Message to `"$DynamicFieldName configured erroneously: $Error"` - in most cases probably not too important, as the "DynamicFieldGet()" would throw an error with the name of the field, but for missing Grid or DF probably more clear.

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -81,28 +81,98 @@ sub new {
...
+    my $DynamicFields = $GetIncludedDynamicFields->($Include, $DynamicFieldObject, $DynamicFieldName);
 
-        if ( !$DynamicField ) {
-            $Kernel::OM->Get('Kernel::System::Log')->Log(
-                Priority => 'error',
-                Message  => "$Param{DynamicFieldConfig}{Name} configured erroneously. 'DF: DynamicFieldName' missing or wrong.",
-            );
+    #DEBUG
+    #use Data::Dumper;
+    #print STDERR "DynamicFields".Dumper($DynamicFields)."\n";
```
svenoe: Please remove.

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -81,28 +81,98 @@ sub new {
...
 
-            return;
-        }
+    for my $i (0 .. $#$DynamicFields) {
```
svenoe: Also here please use `$#{ $DynamicFields }`.

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -130,37 +200,33 @@ sub ValueSet {
...
-                Set                => 1,
-            )
+    my $Include = $Param{DynamicFieldConfig}{Config}{Include};
+    my $DynamicFieldName = $Param{DynamicFieldConfig}{Name};
+    my $DynamicFields = $GetIncludedDynamicFields->($Include, $DynamicFieldObject, $DynamicFieldName);
+
+    if (!$DynamicFields) {
+        
+        $Success = 0;
+    } else {
```
svenoe: Please `return;` instead of `return 0;` or `return undef;` or so, in case of errors. (With the exception of something like `return { Success => 0, Error => 'Did not work' };` where you explicitely want to return more information even in case of an error.)

Here the usual way would be:
```
return if !$DynamicFields;

for my $i ( 0 .. $#{ $DynamicFields } ) {
    ...
```

(One could even argue whether a ValueSet() for a set which has no dynamic fields in it should just `return 1;`, actually, but it's a more theoretical question, I guess...)

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -130,37 +200,33 @@ sub ValueSet {
...
+        
+        for my $i (0 .. $#$DynamicFields) {
+
+            my $DynamicField = $DynamicFields->[$i];
```
svenoe: For my personal taste a bit much, but not wrong ofc.

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -130,37 +200,33 @@ sub ValueSet {
...
+                    %Param,
+                    DynamicFieldConfig => $DynamicField,
+                    Value              => \@FieldValue,
+                    Set                => 1,
+                )
             )
-        {
-            $Success = 0;
+            {
+                $Success = 0;
```
svenoe: The question is whether you even want to try to continue updating the other fields. One maybe could also just return here. Leave a log message though, please. See e.g. https://github.com/RotherOSS/otobo/blob/40a68a17babe7254279c3f27302f037ad8129269/Kernel/System/Ticket.pm#L426-L430
(With an added blank line in front of the return, please.)

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -453,31 +511,31 @@ sub EditFieldValueValidate {
     my $Result;
     for my $i ( 0 .. $#{ $Param{DynamicFieldConfig}{Config}{Include} } ) {
```
svenoe: This looks wrong to me, should have to be deleted, too, I think.

Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -453,31 +511,31 @@ sub EditFieldValueValidate {
...
+        if (!$DynamicFields)
+             {
+            
+            $Result = undef;
+        } else {
+            for my $i (0 .. $#{$DynamicFields}) {
+
+            my $DynamicField = $DynamicFields->[$i];
+                # prevent overwriting names in cached data
+                $DynamicField = { $DynamicField->%* };
```
svenoe: Please combine... ;)

var/httpd/htdocs/skins/Agent/default/css/Core.Form.css
```diff
@@ -707,31 +707,41 @@ fieldset.TableLike > .Row.MultiColumn > .FieldCell > .Field {
...
 }
 
 fieldset.TableLike > .Row > .FieldCell > .Field.DFSetOuterField .DynamicFieldSet label {
+    grid-column: 1;
     padding-top: 0px;
     padding-right: 10px;
     width: 100%;
     box-sizing: border-box;
+    text-align: left;
+    text-overflow: clip;
```
svenoe: Isn't ellipsis nicer?

var/httpd/htdocs/skins/Agent/default/css/Core.InputFields.css
```diff
@@ -345,8 +353,8 @@ div.InputField_ToolbarContainer ul li {
 div.InputField_ToolbarContainer ul li:after {
     color: #bbb;
     content: "|";
-    display: inline-block;
-    float: right;
+    /*display: inline-block;*/ /*inline block is ignored due to float*/
```
svenoe: Then please remove it completely.

var/httpd/htdocs/skins/Customer/default/css/Core.InputFields.css
```diff
@@ -1067,7 +1110,7 @@ div.InputField_ToolbarContainer ul li a:hover {
 
 .jstree-InputField-Tree .jstree-icon {
     font-family: 'otobo' !important;
-    speak: none;
+    /*speak: none; /*unknown property?!*/
```
svenoe: Please research and either leave it in, or remove it completely.

svenoe (changes requested):
In general it looks good, please use the code policy, though (I left some stuff uncommented, where I think, that it will tidy your code a bit anyways). For the Set driver, I did not add comments in places where stuff repeated. In the beginning I wrote quite a lot, also small stuff, just to say as much as possible, further down I looked through the code a bit faster. Especially the css changes are more or less unchecked, as I didn't look at the changes in the frontend. We can make adaptions there, later, too.

Stefan recently added some unit tests specifically for sets. Please ask him which and run them with your changes.

Kernel/Modules/AdminDynamicFieldSet.pm
chzauleck: wouldnt it be nicer to leave the elsif and add error handling in an else clause?

PR #2839, opened 2024-01-17
Add alpine docker image

pat-s:
Working fine so far in a fresh installation.
Build for amd64 and arm64.

Dockerhub link: https://hub.docker.com/repository/docker/cynkra/otobo

Image size alpine: 1.72GB
Image size debian: 2.04GB

Feel free to tweak as needed.

### DIFF
--- a/otobo.web.alpine.dockerfile
+++ b/otobo.web.alpine.dockerfile
@@ -1,0 +1,183 @@
+# This is the build file for the OTOBO web docker image.
+# The services OTOBO web and OTOBO daemon use the same image.
+
+# See also bin/docker/build_docker_images.sh
+# See also https://doc.otobo.org/manual/installation/10.1/en/content/installation-docker.html
+
+# Use the latest maintainance release of the Perl 5.36.x series.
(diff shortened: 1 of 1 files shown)


bschmalhofer:
Hi Patrick,

sorry for the late response. The generous images size was not a priority in the development, but I'm coming back to your PR now. 

For OTOBO 11.1 we make an effort to reduce the image size by switching to  perl:5.40-slim-bookworm as the base image. See #3452. In terms of size reduction this is the same ballpark as switching to alpine. But the step is less radical as still an official Perl image is used. There are also some low hanging fruits when it comes to reducing the image size, by examining which packages really have to be installed.

Once the low hanging fruits are reaped we might examine whether using the system Perl of an image is a good option. This would allow to install the required CPAN modules as OS packages. And this would allow to forgo the installation of compiler tools like gcc.

For now I will apply the PR but then move the new Dockerfile to _development/docker_ as an inspiration for further development.

@pat-s : did you run into any problems running the Alpine base image?

Issue #1884, opened 2022-08-23
Fix Wikipedia Icon in sample Link of ToolBox tile in customer dashboard


bschmalhofer:
The hotlinked icon https://cdn3.iconfinder.com/data/icons/inficons-round-brand-set-2/512/wikipedia-512.png does no longer work. It should be replaced with a working version. I'm also not sure whether hotlinking to the iconfinder CDN is still recommended.

![Bildschirmfoto vom 2022-08-23 13-56-22](https://user-images.githubusercontent.com/46019/186151921-45ee0a89-a962-47f2-9ed7-3855a7351cd4.png)

stefanhaerter:
Is this done? In rel-11_0 it looks like the icon is working again.

bschmalhofer:
An icon recognizable as an icon for Wikipedia is shown. The image is ctually a nice example for a data-url. That's good enough, closing this issue.

PR #2854, opened 2024-01-20
Issue #1884: new icon for Wikipedia in customer interface

bschmalhofer:
Downloaded a new W icon from iconfinder.com as the previous hotlinked icon did no longer exist on the iconfinder CDN. Use the stylized W instead of the puzzleglobe, as the puzzleglobe seems to be the trademark of Wikipedia.
Add the icon as SVG data-url because this way allows to add icons in the SysConfig without uploading any file.


bschmalhofer:
This was discussed briefly in the dev meeting and accepted.

bschmalhofer:
The icon for protonvpn is also broken. For simplicities sake simply remove the sample link, as AFAIK OTOBO has no special connection to protonvpn.

stefanhaerter:

Issue #2855, opened 2024-01-21
Specify an empty import as ()


bschmalhofer:
When an empty list is passed to an `use` statement then the method `import()` of the module is not called. The empty list can be specified as `()` or  as `qw()`. 

```
$ cat t.pl 
use List::Util;
use List::Util qw();
use List::Util ();

$ perl -MO=Deparse t.pl
use List::Util;
use List::Util ();
use List::Util ();
t.pl syntax OK
```

`()` makes a bit more sense as this is the version presented in the documentation of ùse`:  https://perldoc.perl.org/functions/use   . There are also no words to be quoted. So let's standardize on `()` in order to be a bit more consistent. This will also make future tidyings, using  the command perlimports, easier. See also https://perladvent.org/2023/2023-12-23.html .  

The relavant cases can be found with:
`grep -r 'use' | grep 'qw()' | grep -v cpan-lib  | tee ../grep.out `

bschmalhofer:
The test suite still looks fine. PR is merged. Closing this issue.

PR #2856, opened 2024-01-21
Issue #2855: specify an empty import list as ().

bschmalhofer:
As opposed to qw() which is also an empty list, but slightly less tidy.


bschmalhofer:
The test suite looks fine. Merging despite CodePolicy violations.

Issue #2872, opened 2024-01-24
Consistent import of List::Util using the perlimport utility


bschmalhofer:
Just like #2867

bschmalhofer:
PR merged without creating a merge commit. Closing this issue.

PR #2874, opened 2024-01-24
Issue #2872: clean up the 'use' lines for List::Util

bschmalhofer:
in most cases List::Util was not used in the module or script


bschmalhofer:
Starting over

Issue #2894, opened 2024-01-26
Consistent import of Kernel::System::Valid using the perlimport utility 


bschmalhofer:
Run the perlimport utility and clean up the reports about `Kernel::System::ObjectManager`. That module also has no `import()`subroutine,

bschmalhofer:
The was a bit complicated because `use Kernel::System::ObjectManager;` was abused for avoiding "used only once" warnings about `$Kernel::OM`. Made that use case more obvious by adding some ugly `$Kernel::OM = $Kernel::OM` statements.

perlimports is now satisfied about Kernel::System::ObjectManager.  The test suite looks find. PR is applied. Closing this issue.

PR #2895, opened 2024-01-27
Issue #2894 objectmanager

bschmalhofer:


bschmalhofer:
The reported violations appear harmless. Likely  because the CodyPolicy used in the action is not up to date.

Issue #2889, opened 2024-01-26
Fix unit test failures regarding dynamic fields


stefanhaerter:
Currently failing are:

- [x] scripts/test/DynamicField/Backend.t
- [x] scripts/test/DynamicField/EditFieldRender.t
- [x] scripts/test/DynamicField/EditFieldValueGet.t
- [x] scripts/test/Selenium/Customer/DynamicFields/DynamicFieldDatabase.t

stefanhaerter:
Two UT failures fixed with merging #2890

bschmalhofer:
Verified in latest run of the test suite:

> Test Summary Report
> -------------------
> /opt/otobo/scripts/test/Compile.t                                                                         (Wstat: 0 Tests: 1741 Failed: 0)
>   TODO passed:   692, 819
> /opt/otobo/scripts/test/DynamicField/Backend.t                                                            (Wstat: 512 (exited 2) Tests: 131 Failed: 2)
>   Failed tests:  107-108
>   Non-zero exit status: 2
> /opt/otobo/scripts/test/DynamicField/EditFieldValueGet.t                                                  (Wstat: 3072 (exited 12) Tests: 196 Failed: 12)
>   Failed tests:  115-116, 119-120, 123-124, 146-147, 150-151
>                 154-155
>   Non-zero exit status: 12
> /opt/otobo/scripts/test/Selenium/Customer/DynamicFields/DynamicFieldDatabase.t                            (Wstat: 256 (exited 1) Tests: 40 Failed: 1)
>   Failed test:  40
>   Non-zero exit status: 1
> /opt/otobo/scripts/test/Selenium/TestingMethods.t                                                         (Wstat: 0 Tests: 36 Failed: 0)
>   TODO passed:   21, 24, 28
> Files=966, Tests=81436, 4029 wallclock secs (15.70 usr  2.05 sys + 1089.19 cusr 99.74 csys = 1206.68 CPU)
>

bschmalhofer:
Latest run of the test suite:

> Test Summary Report
> -------------------
> /opt/otobo/scripts/test/Compile.t                                                                         (Wstat: 0 Tests: 1741 Failed: 0)
>   TODO passed:   692, 819
> /opt/otobo/scripts/test/DynamicField/EditFieldValueGet.t                                                  (Wstat: 3072 (exited 12) Tests: 196 Failed: 12)
>   Failed tests:  115-116, 119-120, 123-124, 146-147, 150-151
>                 154-155
[...]

PR #2903, opened 2024-01-30
Issue #2889: Re-enact rel-10_1 behavior for checkbox ValueSet with undef value.

stefanhaerter:
Referencing #2889


stefanhaerter:
Closed in favor of #2906

Issue #2916, opened 2024-02-02
Script fields need to be emptied if value is no longer valid on AJAX update


stefanhaerter:
Alike #2911, script fields need to be emptied e.g. if an required argument is no longer present.

stefanhaerter:
Putting this on the wishlist. Decision is to leave this out for the moment and think about it again later.

stefanhaerter:
For clarification: At the moment, this change is deemed to be not useful, therefor adapting label but leaving the branch in case this changes in the future.

stefanhaerter:
As label wontimplement is set, also removing the issue from the wishlist.

PR #2917, opened 2024-02-02
Issue #2916: Set script field value to empty if required arguments are missing which have been present before.

stefanhaerter:
Closing #2916


stefanhaerter:
Decision is to not include this, at least not now.

Issue #2921, opened 2024-02-03
 Consistent import of Kernel::System::DateTime using the perlimport utility


bschmalhofer:
The special case with Kernel::System::DateTime is that this module does export symbols.

```
use Exporter qw(import);

our %EXPORT_TAGS = (    ## no critic qw(OTOBO::RequireCamelCase)
    all => [
        'OTOBOTimeZoneGet',
        'SystemTimeZoneGet',
        'TimeZoneList',
        'UserDefaultTimeZoneGet',
    ],
);

Exporter::export_ok_tags('all');
```
These four method are written in such a way that they ignore their parameter list. So they can be called in several ways:

```
use Kernel::System::DateTime qw(TimeZoneList);

my $ZoneA = TimeZoneList();
my $ZoneB = Kernel::System::DateTime::TimeZoneList();
my $ZoneC =  Kernel::System::DateTime->TimeZoneList;
my $ZoneD =  $DateTimeObject->TimeZoneList;
...
```

bschmalhofer:
Done.

PR #2922, opened 2024-02-03
Issue #2921: perlimports check for Kernel::System::DateTime

bschmalhofer:


bschmalhofer:
Ignoring CodePolicy violations, as the local checks were fine.

Issue #2930, opened 2024-02-06
Introduce light admin functionality


stefanhaerter:

stefanhaerter:
Closed with merging #2930

PR #2931, opened 2024-02-06
Issue #2930: Introduced lightadmin functionality.

stefanhaerter:
Closing #2930


stefanhaerter:
Actually, I overlooked PR #1045 which seems to be very much the same as my changes - I suppose if one PR is chosen, the other one can be closed as well.

Issue #2938, opened 2024-02-07
Bugs in DF Lens


stefanhaerter:
Issue for miscellaneous bugs in DF Lens, will list fixed ones here.

stefanhaerter:
Fixed handling of param ConfigItemHandled, which caused some mess with the cache when referencing a config item from a config item.

stefanhaerter:
Fixed value set behavior of nested lensed by preventing them from using EditFieldValueGet for fetching the referenced object id.

stefanhaerter:
Until now, no other related case has appeared. Closing the issue for now.

PR #2940, opened 2024-02-07
Issue #2938: Nested lens fields

stefanhaerter:
Nested lens fields cannot get their object id from EditFieldValueGet.

Because the reference field of a nested lens field does (usually?) not exist on the mask, in this case the object id has to be retrieved from the database.

Referencing #2938


stefanhaerter:
I made a mistake by thinking that `$LensDFConfig` is the config of the field which the lens points to - therefor closing this PR.

Issue #2945, opened 2024-02-08
Restrict file type of admin package installation to .opm files


stefanhaerter:
By the best of my knowledge, it would make sense to restrict the file type of file upload in admin package installation to .opm files. This would be an enhancement for selecting the file.

stefanhaerter:
> Decision is to not use this because package files can have various other formats, e.g. xslt.

https://github.com/RotherOSS/otobo/pull/2946#issuecomment-1933680833

PR #2946, opened 2024-02-08
Issue #2945: Restrict file type for package installation file upload to opm.

stefanhaerter:
Closing #2945


stefanhaerter:
Decision is to not use this because package files can have various other formats, e.g. xslt.

Issue #2954, opened 2024-02-11
Consistent import of Encode using the perlimport utility


bschmalhofer:
Satisfy perlimports regarding the Encode module.

bschmalhofer:
Done, closing this issue.

PR #2955, opened 2024-02-11
Issue #2954: tidy the perlimports regarding Encode

bschmalhofer:


bschmalhofer:
Ignoring CodePolicy violations as a local run was fine.

Issue #3006, opened 2024-02-21
Fix re-enabling ACL-hidden set fields


stefanhaerter:

stefanhaerter:
Closed with merging #3008

PR #3007, opened 2024-02-21
Issue #3006: Added id to df set html to be found by ACL handling.

stefanhaerter:
Used Activate function instead of InitSelect to also enable inner selects of set fields.

Closing #3006


stefanhaerter:
Closed in favor of #3008

Issue #3010, opened 2024-02-21
Sort dynamic field dropdown and multiselect values by key instead of value


stefanhaerter:

stefanhaerter:
Currently on hold, because if ACLs get into play, BuildSelectionJSON() sorts the values per default, which is inconsistent.

stefanhaerter:
Regarding PR #3011:

> This PR only handles the easy part of the issue, but actually solving it thoroughly requires a lot more work. Closing the PR in the meantime.

The issue is a lot more complex and requires a lot more work. Leaving it for the moment.

PR #3011, opened 2024-02-21
Issue #3010: Sort dropdown values by key instead of value.

stefanhaerter:
Referencing #3010


stefanhaerter:
This PR only handles the easy part of the issue, but actually solving it thoroughly requires a lot more work. Closing the PR in the meantime.

Issue #2842, opened 2024-01-18
FieldRestrictions for new field types.


svenoe:
Move the necessary actions from the FieldRestrictions into the field drivers, include Sets.

PR #3033, opened 2024-02-23
Issue #2842: Fixed iterations over hash keys instead of hash in df set driver.

stefanhaerter:
Referencing #2842


bschmalhofer (commented):
Looks good to me. Sorting the keys might not always be necessary, but it is better to have too much sorting, then miss an instance where sorting is important.

stefanhaerter:
Yes, you are right. I recognized that at approximately half of the occurrences sorting was present, and so I decided to add it everywhere to have a consistent handling.

Issue #3049, opened 2024-02-25
No need to get $LayoutObject in the layout mixin classes


bschmalhofer:
The Perl modules in _ Kernel/Output/HTML/Layout_ seem to be mixed into the LayoutObject. So there is no obvious reason why
a reference to that singleton needs to be fetched from the object manager.

>  grep -r HTML::Layout Kernel/Output/HTML/Layout/*.pm | grep Get
> Kernel/Output/HTML/Layout/LinkObject.pm:    my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
> Kernel/Output/HTML/Layout/Ticket.pm:    my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
> Kernel/Output/HTML/Layout/Translations.pm:    my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
> Kernel/Output/HTML/Layout/Translations.pm:    my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
> Kernel/Output/HTML/Layout/Translations.pm:    my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
> Kernel/Output/HTML/Layout/Znuny4OTOBORepo.pm:    my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');

bschmalhofer:
Cleanup is done, the test suite looks fine. Closing this issue.

PR #3051, opened 2024-02-25
Issue #3049 layout object

bschmalhofer:


bschmalhofer:
Ignoring CodePolicy violations. It looks like Locale::CLDR upgrade is currently on the way.

Issue #3050, opened 2024-02-25
Remove deprecated method Kernel::Language::GetRecommendedCharset and the attribute ReturnCharset


bschmalhofer:
The POD says:

```
=head2 GetRecommendedCharset()

DEPRECATED. Don't use this function any more, 'utf-8' is always the internal charset.

Returns the recommended charset for frontend (based on translation
file or utf-8).

    my $Charset = $LanguageObject->GetRecommendedCharset().

=cut

```

The method is only used to set the attribute `ReturnCharset`. This attribute seems to be unused and can be removed as well.

bschmalhofer:
Done, skipping tests, closing this issue.

PR #3060, opened 2024-02-26
Issue #3050 return charset take2

bschmalhofer:


bschmalhofer:
Ignore problems with CodePolicy, Locale::CLDR currently has problems

Issue #3022, opened 2024-02-22
Remove obsolete attribute 'referrer' from Content-Security-Policy header


bschmalhofer:
In _Kernel/Output/HTML/Layout.pm_ there is a TODO comment. I think that this should be tackled for OTOBO 11.

        # Also disallow referrer headers to prevent referrer leaks via old-style policy directive. Please note this has
        #   been deprecated and will be removed in future OTOBO versions in favor of a separate header (see below).

Details are in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/referrer.

Also the current syntax of the header field looks a bit dubious. Shouldn't the string `no-referrer`be quoted?

```
        $Headers{'Content-Security-Policy'}
            = q{default-src *; img-src * data:; script-src 'none'; object-src 'self'; frame-src 'none'; style-src 'unsafe-inline'; **referrer no-referrer;**};
```

bschmalhofer:
This issue is only about cleaning up the `Content-Security-Policy`header. The recommended header `Referrer-Policy: 'no-referrer'` has already been set since OTOBO 10.0.1.

bschmalhofer:
PR is merged. A quick test was done by creating a dummy phone ticket. When viewing the article of the newly created ticket we get for the request https://localhost:1484/otobo/index.pl?Action=AgentTicketArticleContent;Subaction=HTMLView;TicketID=2;ArticleID=2;FileID=; 

```

HTTP/2 200 
server: nginx/1.25.4
date: Mon, 26 Feb 2024 16:58:37 GMT
content-type: text/html; charset="utf-8"
content-length: 252
content-disposition: inline;  filename="Ticket-2024022610000018-ArticleID-2.html"; filename*=utf-8''Ticket-2024022610000018-ArticleID-2.html
content-security-policy: default-src *; img-src * data:; script-src 'none'; object-src 'self'; frame-src 'none'; style-src 'unsafe-inline';
referrer-policy: no-referrer
x-frame-options: SAMEORIGIN
X-Firefox-Spdy: h2
```

The referret-policy is still at no-referret. The referrer attribute of content-security-policy is gone. Like expected. Closing this issue.

PR #3063, opened 2024-02-26
Issue #3022: remove obsolete attribute 'referrer no-referrer;'

bschmalhofer:
from the Content-Security-Policy header.


bschmalhofer:
Ignore CodePolicy failures because there are problems with Locale::CLDR 0.44.0.

PR #3082, opened 2024-02-28
Update DynamicFields.xml

StefanAbel-OTOBO:
Added description for namespaces to rule out empty spaces. Please make sure no special characters can be used. If that is correct, my description should be correct.


svenoe:
I added a comment plus the actual check in 8156c76dea9ad293580539eeed3adcce508009b8

PR #3083, opened 2024-02-28
Create TicketPrioritySet.pm

StefanAbel-OTOBO:
Please have a good look, if that is correct.
It adds a new Transition Action called "TicketPrioritySet" and requires to enter a Priority name or a Priority ID and additionally allows to override the permissions via a User ID. I tested it on ov215


stefanhaerter (approved):
Seems to be working fine 👍🏻 At first, I stumbled upon priority names being '2 low' instead of 'low', but to the best of my knowlegde, this is the same with all the other transition actions that you have to know the parameters exactly.
Running the CodePolicy brought up two indentation changes and updated the URL in the copyright.

Issue #3087, opened 2024-03-01
Persistent database connections


bschmalhofer:
Currently the database connection does not persist across HTTP requests.

The regular case in OTOBO is that there is a single database connection per process. This connection has either no or a single open statement handle. That statement handle is wrapped by an instance of `Kernel::System::DB`, the DB object. The DB object is a value in the hash `$Kernel::OM` which means that the DB object is destroyed when the object manager is destroyed. The database connections are managed by `DBIx::Connector`. The most important purpose of `DBIx::Connector`is to make sure that there is a DBI database handle available when needed. So `DBIx::Connector` reconnects when a connection is broken. 

The main reason for persistent DB connection is performance.

The groundwork for persistent DB connection is already laid. The DBIx::Connector is cached in a state hash. This means that the connector instance is not destroyed when the object manager is destroyed. But the method `Kernel::System::DB::DESTROY()` explicitly disconnects the database. This can be seen with the MariaDB command `SHOW PROCESSLIST`. Connections from the web container are only present when a request is currently running. As requests are usually fast, this was tested with inserting a `sleep 100:` in _Kernel/Modules/AdminUser.pm_.

So the obvious improvement is to not call `disconnect()` when the DB object is destroyed. Calling `finish()` on the statement handle is fine. However there are three caveats that have to be addressed.

- [x] Can result sets leak from one request to another?
- [x] Are there transactions that are opened in one request that can be still open in the next request?
- [x] Are there database connection that should not be persisted?

bschmalhofer:
Can result sets leak from one request to another?

This would be possible when the handler of one HTTP requests executes a SQL query and does not finish the statement. A statement is finished when either all entries are fetched or when an explicit `finish()`is called on the statement handle. There is an explicit `finish()` called in `Kernel::System::DB::DESTROY()`. This mechanism works reliable, so we don't have to worry about leaked result sets.

It is not obvious whether result sets are stored on the DB server or in the DB client. In the case of OTOBO it is on the DB client, as the default statement handle attribute **mysql_store_result** is set. See https://metacpan.org/pod/DBD::mysql#STATEMENT-HANDLES. But as it is guaranteed that `finish()`is called, the choice between mysql_use_result and mysql_store_result does not matter much.

bschmalhofer:
Are there transactions that are opened in one request that can be still open in the next request?

[...]

PR #3091, opened 2024-03-01
Issue #3087: do not disconnect in cleanup of Kernel::System::DB

bschmalhofer:
This effectively activates DB connections that are persistent over requests. In special cases the old behavior can be restored by passing "DisconnectOnDestruction => 1" to the constructor.


bschmalhofer:
Ignoring CodePolicy violation, as 'Kernel::System::Web::Request' is filed as a soft dependency.

PR #3111, opened 2024-03-04
Load new module for Internal Article Edit and Delete

jacasavilcap:
Load new files and change to existing to enable the article edit and delete feature.


svenoe:
This was all merged locally. Thanks :)

Issue #3146, opened 2024-03-08
Update to CKEditor5


svenoe:
Update to CKEditor5

svenoe:
See commits:
54ed8f054868fc8ccd2dcfdfdfb8480dba1387f4
6033bcb4abd58adef6124deb56d3557b6bd11111
632671ad0ce2343380f600966e3de183e25c66ed
e1ef65f5de49663646c19421b59019819aa08c90
a210a1ca2bc8e57006e54b4996775b5460d9ba82

bschmalhofer:
Adapted _Kernel/Output/HTML/Layout.pm_ in order to avoid a warning about an uninitialized variable during installation. It is fine that there is no RTE during installation, so the warning can simple be avoided.

But there is still an issue with translations in CKEdito. I changed the agents language to Brazilian Purtoguese and the editor was not translated. I think that this is because of a mismatch in the language codes ` pt_BR `vs.` pt-br.

bschmalhofer:
I have been looking inot the current test script failure in _scripts/test/EmailParser/MultipartMixed.t_ . See #3100 . This test is about the parsing of incoming E-mails. The change seems to be that content of _var/httpd/htdocs/common/css/CKEditor.min.css_ has been injected into the incoming HTML attachment. Looks like `Kernel::System::HTMLUtils::DocumentComplete()` is called somewhere. This seems to be fishy. What is the connection between incoming mails and CKEditor?

PR #3161, opened 2024-03-11
Issue #3146: avoid warnings about $BaseEditorPath

bschmalhofer:
when running the OTOBO installer.
Add Frontend::RichTextPath to Kernel::Config::Defaults so that the setting is available when installer.pl is running.


bschmalhofer:
This was nonsense, no need to have a rich text editor during installation.

Issue #2999, opened 2024-02-20
Preserve IncludeInvalid in module subactions


stefanhaerter:
- [x] AdminACL
- [x] Rebuild dynamic fields
- [x] Shift checkbox into own widget

stefanhaerter:
Done, closing the issue.

PR #3204, opened 2024-03-19
Issue #2999: Enable unsetting IncludeInvalid in AdminDynamicField.

bschmalhofer:


bschmalhofer:
Not merging because another adaption of die code_policy action should be tested.

Issue #3213, opened 2024-03-20
Warnings when the Github action code_policy is running


bschmalhofer:
This is not urgent, but should be cleaned up when there is time to spare:

see for example https://github.com/RotherOSS/otobo/actions/runs/8356616927

bschmalhofer:
Most of the warnings are about outdated actions.

The "::set-output" warning is a real change, see https://dev.to/cicirello/how-to-patch-the-deprecated-set-output-in-github-workflows-and-in-container-actions-9co

bschmalhofer:
The last remaining warning is:
```
CodePolicy
The current working directory is not inside a git repository: /__w/otobo/otobo
```

bschmalhofer:
I could not identify what causes that last warning. Let's ignore this warning for the sake of this issue. Closing this issue.

PR #3214, opened 2024-03-20
Issue #3213: switch to tj-actions/changed-files@v43

bschmalhofer:
for getting the list of changed files. This action seems to be better maintained than jitterbit/get-changed-files@v1.


bschmalhofer:
Down to four warnings from code_policy action.

PR #3258, opened 2024-03-27
Article edit/delete features code tidying / fixing

jacasavilcap:
**Tidying:**
- Comparing to Subaction instead of using IsEdit var.
- Moving HistoryAdd blocks to system module ArticleFeatures.pm.

**Fixing:**
- Loop for versions entries is now order ascending.
- Dropdown sort for versions changed to sort by numerical key.


Kernel/Modules/AgentTicketActionCommon.pm
```diff
@@ -376,7 +376,7 @@ sub Run {
     my %ArticleData;
 
     # if we are editing an article
-    if ( $Self->{ArticleID} && !$Self->{IsEdited} ) {
+    if ( $Self->{ArticleID} && !$Self->{Subaction} ) {
```
svenoe, line 379: Just as an example

svenoe:
Cherry picked everything

PR #3260, opened 2024-03-28
Load ckeditor.js only when needed

AjaySDwivedi1:


AjaySDwivedi1:
Not sure if I made the changes in the right places so let me know if I did it wrong or need to update it.

Issue #3223, opened 2024-03-21
Load ckeditor.js only when needed


bschmalhofer:
CKEditor 5 is supported by loading JavaScript libraries in the footer of pages. See https://github.com/RotherOSS/otobo/blob/00268c05ca55726674de03be8c0ecae9df9f365c/Kernel/Output/HTML/Layout.pm#L1810 . Currently that library is loaded for every request. But it would be cleaner to load them only when needed. Another reason is that this would avoid the system call for determining whether the translation file exists.

This could be implemented with the same approach as it is done with the datepicker library. See https://github.com/RotherOSS/otobo/blob/00268c05ca55726674de03be8c0ecae9df9f365c/Kernel/Output/HTML/Layout.pm#L1743

AjaySDwivedi1:
Can I work on this issue? Do I need to change all the if statements that call $ConfigObject->Get() to $HasDatepicker?

bschmalhofer:
Hi @AjaySDwivedi1 ,

you are welcome to work at this and to provide a pull request.

 I suggest that you look at the OTOBO code and understand how the `HasDatePicker` is used. It is basically a flag where frontend modules can indicate that the datepicker JavaScript library is needed for the page. Then you create a similar attribute `HasRichTextEditor` that does the same thing for _ckeditor.js_, Then you have to find all frontends that include a rich text editor and make them raise that new flag.

bschmalhofer:
Sorry @AjaySDwivedi1, I clicked in the wrong tab of my browser and had to revert the PR afterwards.

So your patch is not quite there yet. I fear that solving that issue requires some understanding of OTOBO. So for example see https://github.com/RotherOSS/otobo/blob/df444416feae7c63386f3febf814ca2edf9c4efb/Kernel/Modules/AgentTicketPhone.pm#L3293 . This is an indication that this page uses the rich text editor. So this frontenend module needs to indicate the the _ckeditor.js_ file needs to be loaded. So that at the time the HTML footer is generated, it is know that the library must be included.

So the task is something like this:

- [x] identify the places, that is the frontend modules, where CKEditor is used
- [x] set a flag `HasRichTextEditor`, analoguous to `HasDatepicker`
- [x] add _ckeditor.js_ to the footer only when  `HasRichTextEditor` has been set

But maybe the first step is to get a local development system up and running. See https://doc.otobo.de/manual/installation/11.0/en/content/index.html for inspiration.

AjaySDwivedi1:
[...]

PR #3269, opened 2024-03-29
Load ckeditor.js only when needed in Layout.pm

AjaySDwivedi1:


Kernel/Output/HTML/Layout.pm
```diff
@@ -93,6 +93,7 @@ sub new {
     # allocate new hash for object
     # %Param often has a entry for 'SetCookies'
     my $Self = bless {%Param}, $Type;
+    my $HasRichTextEditor = $Self->{'Frontend::RichText'};
```
bschmalhofer, line 96: The sysconfig setting is the general information that rich text editor is activated. But we need the information specific for a frontend module.

Kernel/Output/HTML/Layout.pm
```diff
@@ -374,14 +375,15 @@ EOF
...
+            Key   => 'RichText',
             Value => 0,
         );
     }
 
     # check if rich text is active
-    if ( !$ConfigObject->Get('Frontend::RichText') ) {
-        $Self->{BrowserRichText} = 0;
+    if (!$HasRichTextEditor) {
+        $Self-> Footer();
```
bschmalhofer, line 385: I dont't understand why Footer() is called here.

Kernel/Output/HTML/Layout.pm
```diff
@@ -1741,7 +1743,7 @@ sub Footer {
     my ( $Self, %Param ) = @_;
 
     my $Type          = $Param{Type}           || '';
-    my $HasDatepicker = $Self->{HasDatepicker} || 0;
+    my $picker = $Self->{HasDatepicker} || 0;
```
bschmalhofer, line 1746: This is a lexical variable. The name `$picker` is perfectly fine in Perl. But the OTOBO coding conventions state that variables should be in camel case.

Kernel/Output/HTML/Layout.pm
AjaySDwivedi1: Would something like $HasRichTextEditor = $Self->{'Frontend::RichText::CKEditor'} be more specific?

bschmalhofer: The specifity is not the issue. OTOBO has a pretty elaborate configuraion system. Take a look at https://github.com/RotherOSS/otobo/blob/f8e9f730af50ee821fbf88f3eff033b706c28604/Kernel/Config/Files/XML/Framework.xml#L376 . The setting Frontend::RichText simply indicates that support for a rich text editor is active. But for all practical purposes, RTE is always enabled.  So this information is not really interesting.

Perhaps it is easier when actually looking at the installation. You can go to https://otobo.de/en/open-source-ticketing-system/#demos and log in as 'Lena/Lena'. You can change the user language at top left. Then you can choose Tickets->New Phone Ticket  to get to a page that has an embedded editor. Looking at the URL you see "Action=AgentTicketPhone". This gives you the information that the frontend module _Kernel/Modules/AgentTicketPhone.pm_ needs to set the flag 'HasRichTextEditor'.

AjaySDwivedi1: Would it be something like this my $HasRichTextEditor = $ConfigObject->Get("Frontend::$Self->{Action}")?

bschmalhofer: No, trial and error is not productive here. 

Perhaps it is better to move away from the actual changes and get a consensus of what should be achieved. Could you go back to #3223 and describe in your own words what should be achieved and how you plan to go ahead with it? In this way we can achieve a common understanding. But please don't despair when you are overwhelmed. OTOBO has an history of over 20 years and things are sometimes a bit peculiar in OTOBO.

bschmalhofer:
I think that a new approach is needed here.

Issue #3305, opened 2024-04-10
Prohibit using same DF across multiple DFSets, TicketMasks and also across TicketMasks and DFSets


stefanhaerter:

stefanhaerter:
Closed with merging #3307

PR #3307, opened 2024-04-10
Issue #3305: Prevent double DF usage

stefanhaerter:
Closing #3305

### DIFF
--- a/Kernel/Modules/AdminDynamicFieldSet.pm
+++ b/Kernel/Modules/AdminDynamicFieldSet.pm
@@ -19,8 +19,6 @@ package Kernel::Modules::AdminDynamicFieldSet;
 use strict;
 use warnings;
 
-use List::Util qw(any);
-
 our $ObjectManagerDisabled = 1;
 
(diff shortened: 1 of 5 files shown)


Kernel/Modules/AdminDynamicFieldSet.pm
```diff
@@ -166,6 +168,19 @@ sub _CheckInclude {
 
     my $DynamicFieldObject        = $Param{DynamicFieldObject};
     my $DynamicFieldBackendObject = $Param{DynamicFieldBackendObject};
+    my $LayoutObject              = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
+    my $TicketMaskObject          = $Kernel::OM->Get('Kernel::System::Ticket::Mask');
+
+    my @Masks = $TicketMaskObject->ConfiguredMasksList();
+    my @MaskDynamicFields;
```
bschmalhofer: As far as I see `@MaskDynamicFields` is used only in a lookup. This would suggest to use a hash like `%DynamicFieldIsUsedInMask`.

Issue #3318, opened 2024-04-17
Add error logging to Kernel::Config::Defaults::AutoloadPerlPackages


stefanhaerter:
Currently, the `require` statement in `sub AutoloadPerlPackages` is placed within a `try` block, but the error case is neither catched nor logged anywhere:

https://github.com/RotherOSS/otobo/blob/fa5e5a92ff1aec72dbc2a492440aea3ad4b2aa7b/Kernel/Config/Defaults.pm#L2399-L2404

It would be nice to get informed if autoloading a module fails.

stefanhaerter:
Closed with merging #3319

PR #3319, opened 2024-04-17
Issue #3318: Added catch block to print error in case of invalid Autoload file.

stefanhaerter:
Error handling derived from comparable occasions within the same file.

Closing #3318


Kernel/Config/Defaults.pm
```diff
@@ -2401,7 +2401,10 @@ sub AutoloadPerlPackages {
                 my $FileName = $Package =~ s{::}{/}smxgr;
 
                 require $FileName . '.pm'; ## nofilter(TidyAll::Plugin::OTOBO::Perl::Require)
-            };
+            }
```
bschmalhofer, line 2404: Looks fine. No need to check the return value of require as require throws an exception in case of problems.

Issue #3329, opened 2024-04-23
Do not remove existing change events when adding FormUpdate events


stefanhaerter:

stefanhaerter:
Closed with merging #3330

PR #3330, opened 2024-04-23
Issue #3329: Do not remove existing change events when adding FormUpdate event.

stefanhaerter:
Closing #3329


bschmalhofer (approved):
Existing handlers should not be removed as they might serve other purposes.

Fix looks fine to me.

Issue #3333, opened 2024-04-24
Fix EditFieldValueValidate in Set DF driver


stefanhaerter:

stefanhaerter:
Closed with merging #3334

PR #3334, opened 2024-04-24
Issue #3333: Fix accessing dynamic field config in param hash.

stefanhaerter:
Closing #3333


bschmalhofer (approved):
Looks like this was not noticed because the effect was that there no validation.

The fix looks good.

Issue #3335, opened 2024-04-25
Make TicketNumber HTML-safe before using it in filenames


stefanhaerter:
While working on a package, I noticed that AgentTicketPrint and CustomerTicketPrint use the plain TicketNumber without sanitisation in their filenames. In most cases the ticket number only consists of digits, which is fine. But because the ticket number can be created by any module put in Kernel/System/Ticket/Number/ and configured in sysconfig setting Ticket::NumberGenerator, I would consider it a nice and low-cost improvement to send it through Kernel::System::Main::FilenameCleanUp().

stefanhaerter:
Closed with merging #3336

PR #3336, opened 2024-04-25
Issue #3335: Clean ticket number before using it in filenames for Ticket printing.

stefanhaerter:
Closing #3335


bschmalhofer (commented):
I suggest to call FilenameCleanup() only after the complet file name has been assembled. This is because there is a length check in FilenameCleanup().

stefanhaerter:
I actually thought about this but thought it would be fine because the other filename components are hardcoded. I simply oversaw the length check - thank you for pointing it out =)

Issue #3338, opened 2024-04-25
Remove accidentally introduced reference to Kernel::System::Time


stefanhaerter:
With issue #1252, Kernel::System::Time was removed for rel-11_0. When adding the translation package, a dependency declaration was accidentally reintroduced.

bschmalhofer:
PR is merged. Closing this issue.

PR #3339, opened 2024-04-25
Issue #3338: Remove dependency to Kernel::System::Time.

stefanhaerter:
While at it, clean up some unused variables.

Closing #3338


bschmalhofer (approved):
Removing not needed variables is good.

Issue #3366, opened 2024-05-03
ImportExport: Add sorting to JSON export


stefanhaerter:
Currently, when exporting data (e.g. ConfigItems) as JSON, the data is unsorted, which leads to different order of hash key / values. This is no problem for most of the time, except when checking the result against a fixed string in unit tests.

However, since the change affects all JSON exports, this needs confirmation.

bschmalhofer:
I second this suggestion. Sorting also makes subsequent exports more comparable.

stefanhaerter:
Provided a PR with the change in question.

bschmalhofer:
This was discussed in the developper meeting and confirmed. PR is merged. Closing this issue.

PR #3367, opened 2024-05-03
Issue #3366: Added sorting to JSON format backend export.

stefanhaerter:
Closing #3366


Kernel/System/ImportExport/FormatBackend/JSON.pm
```diff
@@ -280,6 +280,7 @@ sub ExportDataSave {
     return $JSONObject->Encode(
         Data   => $Param{ExportDataRow},
         Pretty => $Pretty,
+        Sort   => 1,
```
bschmalhofer: The parameter should be 'SortKeys' instead of 'Sort'

stefanhaerter: Fixed, thank you =)

Issue #3407, opened 2024-05-21
Do not try to upgrade packages that are integrated in OTOBO 11.0


bschmalhofer:
A number of OTOBO packages have been integrated into OTOBO 11.0. This means that they do no need to be upgraded when migrating from OTOBO 10 to OTOBO 11. The check for integrated packages should be in the console command `Admin::Package::UpgradeAll`.

l
Here is a sample from the update log:
```
Fetching on-line repositories...
  ITSM Bundle... Done
  ThirdParty Addons... Done
  OTOBO Addons... Done
  OTOpar Addons... Done

Upgrading installed packages...
  The following packages were updated...
    AppointmentToTicket
    DynamicFieldAttachment
    FAQ
    GeneralCatalog
    ITSMCore
    MailAccount-OAuth2
    MasterSlave
  The following packages could not be found in the on-line repositories...
    Ayte-CustomTranslations
    ACME-ServiceCatalog
    ExtendedCDBInfoTile
    ImportExport

Fail.

```
 All but ACME-ServiceCatalog are already part of OTOBO 11 and should not be reported here.

bschmalhofer:
Merged the current adaptions.
TODO:

- [x] Test upgrade from OTOBO 10.0.10 to OTOBO 11.0.1 again

bschmalhofer:
Tested the upgrade from 10.1 to 11.0 in a Docker based installation. The output in that test looks like:

```
Fetching on-line repositories...
787   ITSM Bundle... Done
788   ThirdParty Addons... Done
789   OTOBO Addons... Done
790   OTOpar Addons... Done
791 
792 Upgrading installed packages...
793   The following packages were updated...
794     AppointmentToTicket
795     DownloadAllAttachments
796     GeneralCatalog
797     ITSMCore
798     ImportExportCustomerCompany
799     ImportExportTicket
800   The following packages could not be found in the on-line repositories...
801     ITSMChangeManagement
802     ITSMConfigurationManagement
803     ITSMIncidentProblemManagement
804     ITSMServiceLevelManagement
805 
806 Fail.
```
The packages ImportExport and ExtendedCDBInfoTile are not reported as they were skipped. Looks like it works as intended. Closing this issue.

laiminhtrung1997:
Dear @bschmalhofer 
How can I install package ITSM in OTOBO 11.0.6? There is no version 11.0.x for ITSM.

svenoe:
[...]

PR #3411, opened 2024-05-21
Issue #3407 upgrade all

bschmalhofer:


bschmalhofer:
Ignoring CodePolicy violations. This was only about useless 'no critic' declarations.

Issue #3100, opened 2024-03-04
Fix UnitTest Failures


stefanhaerter:
- [x] EditFieldRender
- [x] Escalations
- [x] AgentTicketFreeText
- [x] CustomerTicketZoom-related
- [x] AgentTicketBulk(ACL)
- [x] AgentTicketForward(Draft)
- [x] AgentTicketEmailOutbound
- [x] AgentTicketEmailPGP
- [x] AgentTicketEmail(Draft)
- [x] Agent/InlineTicketImages
- [x] ProcessManagement/Bug12179
- [x] Admin/TicketAppointments

bschmalhofer:
Things look better now. Here is the result of the latest run, with "local in Eventhandler commented out":

> Test Summary Report
> -------------------
> /opt/otobo/scripts/test/Compile.t                                                                         (Wstat: 0 Tests: 1775 Failed: 0)
>   TODO passed:   695, 825
> /opt/otobo/scripts/test/Selenium/Agent/AgentTicketActionCommon/AgentTicketFreeText.t                      (Wstat: 2304 (exited 9) Tests: 64 Failed: 9)
>   Failed tests:  38, 40, 43, 46, 48, 50, 52, 54, 56
>   Non-zero exit status: 9
> /opt/otobo/scripts/test/Selenium/TestingMethods.t                                                         (Wstat: 0 Tests: 36 Failed: 0)
>   TODO passed:   21, 24, 28
> Files=968, Tests=81640, 4046 wallclock secs (15.99 usr  2.16 sys + 1093.35 cusr 102.84 csys = 1214.34 CPU)
> Result: FAIL
>

stefanhaerter:
Thank you very much =)
`scripts/test/Selenium/Agent/AgentTicketActionCommon/AgentTicketFreeText.t` has a known issue, most likely related to focus handling and my recent FormUpdate changes. I will investigate it as soon as I have the time.
The fact that the tests `scripts/test/Event/Escalations.t`, `scripts/test/ProcessManagement/Bug12179.t` and `scripts/test/Selenium/Agent/Admin/TicketAppointments.t` do not fail indicates that they are somehow related to the EventHandler changes, I guess.

bschmalhofer:
The latest test run with EventHandler.pm as checked in:

> 
> Test Summary Report
> -------------------
> /opt/otobo/scripts/test/Compile.t                                                                         (Wstat: 0 Tests: 1779 Failed: 0)
[...]

PR #3442, opened 2024-06-05
Issue #3100: Updated EditFieldRender test to respect recent reference driver changes.

stefanhaerter:
Referencing #3100

### DIFF
--- a/scripts/test/DynamicField/EditFieldRender.t
+++ b/scripts/test/DynamicField/EditFieldRender.t
@@ -2,7 +2,7 @@
 # OTOBO is a web-based ticketing system for service organisations.
 # --
 # Copyright (C) 2001-2020 OTRS AG, https://otrs.com/
-# Copyright (C) 2019-2024 Rother OSS GmbH, https://otobo.de/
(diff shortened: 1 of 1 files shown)


scripts/test/DynamicField/EditFieldRender.t
```diff
@@ -79,18 +79,27 @@ my $SecondUserID = $UserObject->UserAdd(
...
-    )
+    );
+    my %Preferences = $UserObject->GetPreferences(
+        UserID => $_
+    );
+    $_ => $LayoutObject->Ascii2Html( Text => qq{"$UserName" <$Preferences{UserEmail}>} )
 } keys %UserList;
 
-my $UserSelectionString               = '  <option value="">-</option>';
+my $UserSelectionString;
```
bschmalhofer, line 102: This should be initialized with an empty string in order to avoid a warning later.

stefanhaerter: Ah, hm... perhaps you're right. I encountered the problem that the join later on places a newline at the beginning, but I had to remove this anyways by regex, so I think empty string initialization is fine. I will change this, thank you!

Issue #3424, opened 2024-05-27
Have a look at the sorting of articles together with the article edit/delete feature.


svenoe:
I'm wondering whether in that SQL an ORDER BY clause is missing: https://github.com/RotherOSS/otobo/blob/5de8569a76ab70a3ba842738eec67a766710e370/Kernel/System/Ticket/Article.pm#L1225-L1233
This could be related to some failures in the test suite.

_Originally posted by @bschmalhofer in https://github.com/RotherOSS/otobo/issues/3157#issuecomment-2036377073_

stefanhaerter:
According to https://forums.mysql.com/read.php?21,239471,239688#msg-239688, I would strongly suggest to add an order clause in the above sql.

> Do not depend on order when ORDER BY is missing.

Also, the `ORDER BY a.id ASC` in

https://github.com/RotherOSS/otobo/blob/2972ad1ab92a9aaa1db49c30f197b6edf2957f0b/Kernel/System/Ticket/Article.pm#L1221

and the `ORDER BY at.create_time ASC` in

https://github.com/RotherOSS/otobo/blob/2972ad1ab92a9aaa1db49c30f197b6edf2957f0b/Kernel/System/Ticket/Article.pm#L1245

are contradictory as the first will give something like

```
* ---------- * ----------- *
| article_id | create_time |
* ---------- * ----------- *
| 1          | 2024-01-01  |
| 2          | 2024-01-02  |
| 3          | 2024-01-03  |
* ---------- * ----------- *
```

while the second will produce something like

```
* ---------- * ----------- *
| article_id | create_time |
* ---------- * ----------- *
| 3          | 2024-01-03  |
| 2          | 2024-01-02  |
| 1          | 2024-01-01  |
* ---------- * ----------- *
```

if I understand the following source correctly:

> Same rules apply for the letters of the alphabet. When arranging them in ascending order they are arranged from A to Z – or beginning to end.
> 
> When it comes to dates, ascending order would mean that the oldest ones come first and the most recent ones last.

https://www.freecodecamp.org/news/descending-order-vs-ascending-order-what-does-it-mean/

stefanhaerter:
Should be done with merged #3457

PR #3457, opened 2024-06-10
Issue #3100: Added sorts to ArticleList sql where it seemed necessary.

stefanhaerter:
Extended cache key for ArticleList.

Referencing #3100 and closing #3424

### DIFF
--- a/Kernel/System/Ticket/Article.pm
+++ b/Kernel/System/Ticket/Article.pm
@@ -1197,13 +1197,16 @@ sub _MetaArticleList {
         return;
     }
 
+    my $ShowDeletedArticles = $Param{ShowDeletedArticles} ? 1 : 0;
+    my $VersionView         = $Param{VersionView}         ? 1 : 0;
+
     my $CacheKey
         = '_MetaArticleList::'
         . $Param{TicketID}
         . '::ShowDeletedArticles::'
-        . int( $Param{ShowDeletedArticles} || 0 )
+        . $ShowDeletedArticles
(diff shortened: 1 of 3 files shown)


Kernel/System/Ticket/Article.pm
```diff
@@ -1197,17 +1197,21 @@ sub _MetaArticleList {
         return;
     }
 
-    my $CacheKey = '_MetaArticleList::' . $Param{TicketID};
+    my $CacheKey
```
bschmalhofer, line 1203: Just a bit nitpicking. For VersionView and ShowDeletedArticles any true value could be passed. But the later cache clearing code only clears for the values 0 and 1. This edge case could be avoided by using dedicated variables 
`my $VersionView = $Param{VersionView} ? 1 : 0;`

stefanhaerter: I intended to handle this with `int( $Param{VersionView} )` - is this not sufficient? And if so, why?

bschmalhofer: If hypothetically `VersionView => 99.9` is passed, than int('99.9') = 99 would be used as Key for getting and setting the cache. But 1 would still be the Key for clearing the cache.

stefanhaerter: I see - you're right, as usual ;) I'll change this

bschmalhofer (approved):
Looks good to me.

Issue #3537, opened 2024-06-26
Enable HideShow for AgentTicketCompose


stefanhaerter:

stefanhaerter:
PR #3633 merged, closing the issue.

PR #3539, opened 2024-06-26
Issue #3537: Enabling hide / show for AgentTicketCompose.

stefanhaerter:
Closing #3537

### DIFF
--- a/Kernel/Modules/AgentTicketCompose.pm
+++ b/Kernel/Modules/AgentTicketCompose.pm
@@ -2516,6 +2516,14 @@ sub _Mask {
     my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
     my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
 
+    if ( $Param{HideAutoselected} ) {
+
+        # add Autoselect JS
+        $LayoutObject->AddJSOnDocumentComplete(
+            Code => "Core.Form.InitHideAutoselected({ FieldIDs: $Param{HideAutoselected} });",
+        );
+    }
+
     # get config for frontend module
     my $Config = $ConfigObject->Get("Ticket::Frontend::$Self->{Action}");
 
--- a/Kernel/Modules/AgentTicketCompose.pm
+++ b/Kernel/Modules/AgentTicketCompose.pm
@@ -1376,6 +1376,9 @@ sub Run {
         my @DynamicFieldAJAX;
 
         my $ElementChanged = $ParamObject->GetParam( Param => 'ElementChanged' ) || '';
+        if ( $ElementChanged eq 'StateID' ) {
+            $ElementChanged = 'NextStateID';
+        }
 
         # get list type
         my $TreeView = 0;
@@ -1477,19 +1480,19 @@ sub Run {
                 if ( $GetParam{StateID} && !$NextStates->{ $GetParam{StateID} } ) {
 
                     # if not empty the field
-                    $GetParam{StateID}           = '';
-                    $NewChangedElements{StateID} = 1;
-                    $ChangedStdFields{StateID}   = 1;
+                    $GetParam{StateID}               = '';
+                    $NewChangedElements{NextStateID} = 1;
+                    $ChangedStdFields{NextStateID}   = 1;
                 }
 
                 # autoselect
-                elsif ( !$GetParam{StateID} && $Autoselect && $Autoselect->{StateID} ) {
+                elsif ( !$GetParam{StateID} && $Autoselect && $Autoselect->{NextStateID} ) {
                     $GetParam{StateID} = $FieldRestrictionsObject->Autoselect(
                         PossibleValues => $NextStates,
                     ) || '';
                     if ( $GetParam{StateID} ) {
-                        $NewChangedElements{StateID} = 1;
-                        $ChangedStdFields{StateID}   = 1;
+                        $NewChangedElements{NextStateID} = 1;
+                        $ChangedStdFields{NextStateID}   = 1;
                     }
                 }
 
@@ -1534,7 +1537,8 @@ sub Run {
                     CustomerUser              => $Selected || '',
                     GetParam                  => {
                         %GetParam,
-                        OwnerID => $GetParam{NewUserID},
+                        NextStateID => $GetParam{StateID},
+                        StateID     => undef,
                     },
                     Autoselect      => $Autoselect,
                     ACLPreselection => $ACLPreselection,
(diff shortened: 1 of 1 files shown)


Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -86,19 +86,19 @@ sub new {
...
+        $Self->{DynamicField}{$DynamicFieldName} = $DynamicFieldObject->DynamicFieldGet(
             Name => $DynamicFieldName,
         );
 
         my $Parameters = $Definition->{DynamicFields}{$DynamicFieldName} // {};
 
         for my $Attribute ( keys $Parameters->%* ) {
-            $Self->{DynamicField}[-1]{$Attribute} = $Parameters->{$Attribute};
+            $Self->{DynamicField}{$DynamicFieldName}{$Attribute} = $Parameters->{$Attribute};
         }
```
svenoe, line 102: Please align the code more with the main masks, AgentTicketActionCommon would be the most general one, even though some things, like the differentiation of ticket and article dynamic fields are not necessary here, of course.

In the new method therefor please also define $Self->{FieldMethods} (only containing the method for the next state), as well as $Self->{InternalDependancy} (as empty hash).

Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -473,7 +474,7 @@ sub Run {
 
     # cycle trough the activated Dynamic Fields for this screen
     DYNAMICFIELD:
-    for my $DynamicFieldConfig ( @{ $Self->{DynamicField} } ) {
+    for my $DynamicFieldConfig ( values $Self->{DynamicField}->%* ) {
         next DYNAMICFIELD if !IsHashRefWithData($DynamicFieldConfig);
 
         # extract the dynamic field value from the web request
```
svenoe, line 480: Prio not high, but the lines below `# convert dynamic field values into a structure for ACLs` (490ff - I cannot "review" there, exactly) could be aligned e.g. with ActionCommon.

Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -841,13 +842,117 @@ sub Run {
...
+        # skip validation of hidden fields
+        my %Visibility;
+
+        # transform dynamic field data into DFName => DFName pair
+        my %DynamicFieldAcl = map { $_ => $_ } keys $Self->{DynamicField}->%*;
+
+        # call ticket ACLs for DynamicFields to check field visibility
+        my $ACLResult = $TicketObject->TicketAcl(
+            %GetParam,
+            CustomerUserID => $Ticket{CustomerUser} || '',
```
svenoe, line 854: This differs from ATAC, not sure by heart how it is supposed to be, though. (The same at similar places further down.)

Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -841,13 +842,117 @@ sub Run {
...
+
+                # propagate validation error to the Error variable to be detected by the frontend
+                if ( $ValidationResult->{ServerError} ) {
+                    $Error{ $DynamicFieldConfig->{Name} }                        = ' ServerError';
+                    $DynamicFieldValidationResult{ $DynamicFieldConfig->{Name} } = $ValidationResult;
+                }
+            }
+        }
+
         # cycle trough the activated Dynamic Fields for this screen
```
svenoe, line 953: Isn't this whole loop exactly duplicate from the above?

Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -1094,7 +1205,7 @@ sub Run {
         # set dynamic fields
         # cycle trough the activated Dynamic Fields for this screen
         DYNAMICFIELD:
-        for my $DynamicFieldConfig ( @{ $Self->{DynamicField} } ) {
+        for my $DynamicFieldConfig ( values $Self->{DynamicField}->%* ) {
             next DYNAMICFIELD if !IsHashRefWithData($DynamicFieldConfig);
```
svenoe, line 1209: Be careful, we have to skip hidden fields here.

Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -1264,60 +1375,316 @@ sub Run {
         # update Dynamic Fields Possible Values via AJAX
```
svenoe, line 1375: This time above: We get the next states within the convergence loop now, below, so the above can be deleted.

Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -1264,60 +1375,316 @@ sub Run {
...
+                                next FIELD;
+                            }
+                        }
+
+                        # delete unaffected fields
+                        delete $Check{$FieldID};
+                    }
+                }
+
+                # check value of state
```
svenoe, line 1466: If you define the FieldMethods above, you should be able to almost 1:1 copy this whole section from any of the other modules, making it easier to maintain. Just %Check has to be defined individually, because we only have the NextStateID here.

Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -1264,60 +1375,316 @@ sub Run {
...
+                    DynamicFields             => $Self->{DynamicField},
+                    DynamicFieldBackendObject => $DynamicFieldBackendObject,
+                    ChangedElements           => \%ChangedElements,            # optional to reduce ACL evaluation
+                    Action                    => $Self->{Action},
+                    UserID                    => $Self->{UserID},
+                    FormID                    => $Self->{FormID},
+                    CustomerUser              => $Selected || '',
+                    GetParam                  => {
+                        %GetParam,
+                        NextStateID => $GetParam{StateID},
```
svenoe, line 1532: When aligning the things, NextStateID should already be used in the code, so no need for this anymore, too.

Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -1264,60 +1375,316 @@ sub Run {
...
+            if (%ChangedElements) {
+
+                # get values and visibility of dynamic fields
+                %CurFieldStates = $FieldRestrictionsObject->GetFieldStates(
+                    TicketObject              => $TicketObject,
+                    DynamicFields             => $Self->{DynamicField},
+                    DynamicFieldBackendObject => $DynamicFieldBackendObject,
+                    ChangedElements           => \%ChangedElements,            # optional to reduce ACL evaluation
+                    Action                    => $Self->{Action},
+                    UserID                    => $Self->{UserID},
```
svenoe, line 1527: Please add the TicketID which is important for PropertiesDatabase in the ACLs.

Kernel/Modules/AgentTicketCompose.pm
```diff
@@ -1813,59 +2180,244 @@ sub Run {
...
 
-            my $PossibleValuesFilter;
+        # gather fields which are supposed to be hidden when autoselected
+        my $HideAutoselectedJSON;
+        if ($Autoselect) {
+            my @HideAutoselected = grep { !ref( $Autoselect->{$_} ) && $Autoselect->{$_} == 2 } keys %{$Autoselect};
+            AUTOSELECT:
+            for my $Item (@HideAutoselected) {
+                next AUTOSELECT if $Item ne 'NextStateID';
+                $Item = 'StateID';
```
svenoe, line 2198: Please also align the below (not making comments there, see AJAXUpdate), and changing this here should not be necessary then.

svenoe (changes requested):

stefanhaerter:
Closed in favor of #3633

Issue #3617, opened 2024-07-17
Use pinned versions of CPAN distributions for rebuilding Docker images


bschmalhofer:
> When the check for changes in the base images is implemented then there will be fairly frequent rebuilds of Docker images. 
> These rebuilds include the installation of CPAN distributions into the Docker image. Currently most of the CPAN dependencies are installed with their most current versions on CPAN. This is fine for infrequent builds as there the builds are tested before they are tagged and built automatically on Docker Hub. For rebuilding already released versions we want a more stable setup. Specifically we want that the same versions of CPAN distrubutions  are installed into the rebuilt image. This can be achieved with _cpanfile.snapshot_ support in [Carton](https://metacpan.org/release/MIYAGAWA/Carton-v1.0.35).
> 
> See also #210.

_Originally posted by @bschmalhofer in https://github.com/RotherOSS/otobo/issues/3448#issuecomment-2233043883_

bschmalhofer:
The inital approach is that local build update _cpanfile,snapshot_ and that the automated builds use it.

TODO:

- [x] enhance _bin/docker/build_docker_images.sh_ to copy  _cpanfile.snapshot_  from the image to the git sandbox
- [x] add _cpanfile.snapshot_ , or maybe _cpanfile.docker.snapshot_,  to the git repository
- [x] tell the automated builds on hub.docker.com to use `carton install --deployment`
- [x] decide whether the rebuilds are done on Github or on Docker Hub

bschmalhofer:
The automated builds initially did not accept the heredoc syntax in the Dockerfile _otobo.web.dockerfile_ . Declaring the Dockerfile frontend in the Dockerfile did the trick.

```
$ more otobo.web.dockerfile -n 5
# syntax=docker/dockerfile:1.9

# This is the build file for the OTOBO web docker image.
# The services OTOBO web and OTOBO daemon use the same image.
# There is also an extra build target otobo-web-kerberos that adds support for Kerberos.

```

bschmalhofer:
There still is a problem. When changes from one branch are merged the next higher version branch, e.b. rel-10_1 merged into rel-11_0, then _cpanfile.docker.snapshot_ would also be merged. Excluding files from `git merge` is not trivially possible.

bschmalhofer:
[...]

PR #3618, opened 2024-07-17
Issue #3617 cpanfile snapshot

bschmalhofer:


bschmalhofer:
Looks like Docker Hub is not happy with the heredoc syntax. The automatic build fails with:

> ... dockerfile parse error line 105: unknown instruction: SET

Issue #3621, opened 2024-07-18
Create custom reply method for indicating variable status codes to JS


stefanhaerter:

stefanhaerter:
Issue closed with merged PR.

PR #3627, opened 2024-07-18
Issue #3621: Provide custom exception method to indicate error to AJAX.

stefanhaerter:
Closing #3621

### DIFF
--- a/Kernel/Output/HTML/Layout.pm
+++ b/Kernel/Output/HTML/Layout.pm
@@ -2936,20 +2936,16 @@ Throws an exception with customizable status code and body. Status can either be
 sub AJAXException {
     my ( $Self, %Param ) = @_;
 
-    # if given, check validity of status code
+    # check validity of status code
     # allowed are client errors (400-499) and server errors (500-599)
-    my $StatusCode;
-    if ( $Param{StatusCode} ) {
-        if ( !is_client_error( $Param{StatusCode} ) && !is_server_error( $Param{StatusCode} ) ) {
-            $Kernel::OM->Get('Kernel::System::Log')->Log(
-                Priority => 'error',
-                Message  => "Status code $Param{StatusCode} is not a valid HTTP error status!",
-            );
-            return;
-        }
-        $StatusCode = $Param{StatusCode};
+    my $StatusCode = $Param{StatusCode} || 500;
+    if ( !is_client_error($StatusCode) && !is_server_error($StatusCode) ) {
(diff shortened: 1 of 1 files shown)


Kernel/Output/HTML/Layout.pm
```diff
@@ -2919,6 +2920,50 @@ sub JSONReply {
...
+    my ( $Self, %Param ) = @_;
+
+    # if given, check validity of status code
+    # allowed are client errors (400-499) and server errors (500-599)
+    my $StatusCode;
+    if ( $Param{StatusCode} ) {
+        if ( !is_client_error( $Param{StatusCode} ) && !is_server_error( $Param{StatusCode} ) ) {
+            $Kernel::OM->Get('Kernel::System::Log')->Log(
+                Priority => 'error',
+                Message  => "Status code $Param{StatusCode} is not a valid HTTP status!",
```
bschmalhofer: The message could be more specific "is not an error HTTP status"

bschmalhofer (approved):
Looks good to me

Kernel/Output/HTML/Layout.pm
stefanhaerter: Thank you :) Changed this.

Kernel/Output/HTML/Layout.pm
```diff
@@ -2919,6 +2920,50 @@ sub JSONReply {
...
+        if ( !is_client_error( $Param{StatusCode} ) && !is_server_error( $Param{StatusCode} ) ) {
+            $Kernel::OM->Get('Kernel::System::Log')->Log(
+                Priority => 'error',
+                Message  => "Status code $Param{StatusCode} is not a valid HTTP error status!",
+            );
+            return;
+        }
+        $StatusCode = $Param{StatusCode};
+    }
+    $StatusCode ||= 500;
```
bschmalhofer: Being a bit nitpickey here, I suggest to write the assignment as
`my $StatusCode = $Param{StatusCode} || 500;`
The reasoning is that variables should have limited scope.

stefanhaerter: But how do you then access the variable for usage in `Plack::Response->new()`, if you scope it to the if-Block?

stefanhaerter: What do you think about the updated version I just pushed?

bschmalhofer: Excellent, that is better than mine. First handle the default case and then the sanity check.

Issue #3532, opened 2024-06-25
Custom translations don't work correctly when containing '::'


stefanhaerter:
E.g. when adding custom translations for hierarchical structures like queues or services, translations like *Service1::Service1.1::Service1.1.1* will cause a wrong sorting in dropdown lists and thus destroy the tree structure, as it is based on the correct sorting of the elements.

TODO:

- [ ] Review sorting mechanism 
- [ ] Display source strings in a way which makes it clear that only the last element should be translated
- [ ] Adapt the custom translation package to mirror the changes neccessary

stefanhaerter:
Bug couldn't be reproduced, should be fixed with commit https://github.com/RotherOSS/otobo/commit/529d94214df8c09013d4aacd20f9a38c7ce3a941.

Closing the issue.

PR #3653, opened 2024-07-30
Issue #3532: Added generation of automatic chained translations for queues and services.

stefanhaerter:
Displayed queues and services without its parents in translation tables.

Referencing #3532


stefanhaerter:
Closed in favor of #3659

Issue #3660, opened 2024-08-01
Refactoring using List::Util::pairs


stefanhaerter:
I assume that in at least a few places iterations over a hash are done where key and value are processed. In this places, [List::Util::pairs](https://metacpan.org/pod/List::Util#pairs) could be a viable refactoring option.

stefanhaerter:
Putting this on the wishlist and closing it.

stefanhaerter:
Provided a small example in #3671

bschmalhofer:
I think that we won't do a full refactoring where all loops over hashes are changed. So I think that having an example is good enough for now. New code can then make use of `List::Util::pairs()`.

stefanhaerter:
Agree, sounds like a good idea.

PR #3671, opened 2024-08-01
Issue #3660: Used List::Util::pairs in DF Set driver where possible.

stefanhaerter:
Referencing #3660

### DIFF
--- a/Kernel/System/DynamicField/Driver/Set.pm
+++ b/Kernel/System/DynamicField/Driver/Set.pm
@@ -27,6 +27,7 @@ use utf8;
 use parent qw(Kernel::System::DynamicField::Driver::Base);
 
 # core modules
+use List::Util qw(pairs);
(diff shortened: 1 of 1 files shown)


Kernel/System/DynamicField/Driver/Set.pm
```diff
@@ -419,8 +420,8 @@ sub EditFieldValueGet {
             $DynamicField->{$Name} = { $DynamicField->{$Name}->%* };
         }
 
-        for my $Name ( sort keys $DynamicField->%* ) {
-            my $DynamicFieldConfig = $DynamicField->{$Name};
+        for my $Pair ( pairs $DynamicField->%* ) {
```
bschmalhofer, line 423: An alternative name for `$Pair` could be `$KeyValue`, which would make the semantics more clear.

PR #3695, opened 2024-08-09
Example PR

stefanhaerter:
Provide example of Perl::Critic policies RegularExpressions::RequireDefault and Variables::ProhibitUnusedVarsStricter.


stefanhaerter:
Learning: It is probably not that nice to require `/a` everywhere even if it wouldn't be neccessary. Postponing the topic and closing the PR.

Issue #3697, opened 2024-08-09
Improve RegExp safety


stefanhaerter:

stefanhaerter:
Duplicate of #3038

bschmalhofer:
The problems are mostly with character classes and with lower and upper case. See for example https://perldoc.perl.org/perlre#/a-(and-/aa) .

PR #3698, opened 2024-08-09
Issue #3697: Replace \d with [0-9] in Kernel/GenericInterface.

stefanhaerter:
Referencing #3697


bschmalhofer (approved):
I'm pretty sure that it was never intended that non-arabic numerals should match in these regexes. LGTM.

Issue #3682, opened 2024-08-06
Add console commands for Export, Import.  Including chunking. Improve Ticket import.


bschmalhofer:
ImportExport is integrated in OTOBO 11.0 and OTOBO 11.1. But there are no console commands that perform Import and Export. So let's provide them. The new commands will be almost exact copies of the commands in ITSMConfigurationManagement, https://github.com/RotherOSS/ITSMConfigurationManagement/tree/master/Kernel/System/Console/Command/Admin/ITSM/ImportExport .

Support for chunking will be added to the commands and to the Translations object backed. Chunking the translations is basically only meant as an example, as it is unlikely that translation will exhaust the machines memory.

Do not add the commands in rel-11_0, as patch releases should have no new features.

bschmalhofer:
Tentatively added the commands. 

TODO:

- [x] expand test scripts to actually export and import translations
- [x] support chunked export 
- [x] support chunked import
- [x] make final decision on the naming conventions, @svenoe 
- [ ] manual test

Additional fixes to Import Export:

- [x] Do not count the writing of the header line as a successful export
- [x] add a line break after the last line of the export

bschmalhofer:
The current state is available at https://github.com/RotherOSS/otobo/tree/issue-%233682-importexport_with_chunking.

bschmalhofer:
Another thing to discuss:

The new files _Kernel/System/Console/Command/Admin/ImportExport/Import.pm_ and _Kernel/System/Console/Command/Admin/ImportExport/Export.pm_ are planned to be added to OTOBO 11.1. For OTOBO 11.0 the plan is to backport these files to the [ImportExportTicket](https://github.com/RotherOSS/ImportExportTicket) package. 

It can be considered to integrated the two files from ImportExportTicket into OTOBO 11.1.

```
<Filelist>
        <File Permission="660" Location="Kernel/Config/Files/XML/ImportExportTicket.xml" />
        <File Permission="660" Location="Kernel/System/ImportExport/ObjectBackend/Ticket.pm" />
</Filelist>
```

This would allow to mark ImportExportTicket simply as integrated, without having to consider the two different cases.

bschmalhofer:
Discussed this with @svenoe .

- The two files from the ImportExportTicket package will be integrated into rel-11_1
- The console commands get the new option `prefix` stating the prefix of the files, default is the empty string

bschmalhofer:
[...]

PR #3710, opened 2024-08-14
Issue #3682: integrate the package ImportExportTicket

bschmalhofer:
actually only one new file, the config has been integrated in existing XML file


bschmalhofer:
Ignoring complaint about useless use of `## no critic`.

PR #3731, opened 2024-08-27
Efflux: Disable logging of plaintext passwords in debugging mode (CVE-2024-43444)

eyazi:
Disable the logging of plaintext passwords in debugging mode.

This is inspired by CVE-2024-43444. I'd not consider this a security patch, as the debugging mode can only be enabled by changing the source code. And if you have access to the source code, you can debug anything.

### DIFF
--- a/Kernel/System/CustomerAuth/DB.pm
+++ b/Kernel/System/CustomerAuth/DB.pm
@@ -122,6 +122,7 @@ sub Auth {
     my $RemoteAddr  = $ParamObject->RemoteAddr() || 'Got no REMOTE_ADDR env!';
     my $UserID      = '';
     my $GetPw       = '';
+    my $Method      = '';
 
     # sql query
     $Self->{DBObject}->Prepare(
@@ -156,6 +157,7 @@ sub Auth {
 
     if ( $Self->{CryptType} eq 'plain' ) {
         $CryptedPw = $Pw;
+        $Method    = 'plain';
     }
 
     # md5 or sha pw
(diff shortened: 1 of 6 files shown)


bschmalhofer:
I would basically leave the code as it is. Debug mode is meant to be activated only by making a change in the code of `new()`. This is a safeguard for inadvertently logging passwords. Howerver, this approach could be documented in a better way.

The attribute `Debug` could also be set from the outside by calling `$Auth->{Debug} = 1` but then an attacker would not need to look in the log anyways.

The crypt method `plain`just says that passwords are stored in plain text in the database, which is a bad practice. But that does not imply that these passwords are less sensible.

eyazi:
@bschmalhofer Thanks for the fast reply.

 I’ve made some changes to make the PR more consistent.

In my understanding of security, an application should never write passwords in plain text to a log file. If crypt type is set to plain, for whatever reason, even passwords retrieved from the database are logged in plain text.
However, I understand that there might be different perspectives on this, especially since you need to set the minimum log level to notice and enable the logging directly in the code.

Kernel/System/CustomerAuth/DB.pm
```diff
@@ -156,6 +156,7 @@ sub Auth {
 
     if ( $Self->{CryptType} eq 'plain' ) {
         $CryptedPw = $Pw;
+        $Method    = 'plain';
```
bschmalhofer, line 160: $Method is not declared

bschmalhofer:
Sorry Emin,
I totally misread your first PR, reading 'eq' when 'ne' was written. I thought that you meant that it is OK to dump the plain text password. Could you fix the compile error with `$Method` in _Kernel/System/Auth/DB.pm_ ? I'll apply the PR then.

Actually I think that not even the crypted passwords should be logged. I see no use case for logging passwords, besides deep debugging. I think that should be discussed with @svenoe .

eyazi:
Thanks for noticing that, Bernhard. I’m not sure how I missed it.

I created this PR to address the CVE without thinking too much about improvements. But I doubt that too many people actually enable logging via the source code. Displaying the salted password could likely be removed as well, but in that case, you might consider removing the entire debugging code.

bschmalhofer (approved):
Looks good, after applying the patch I will replace ` $Self->{Debug}` with lexical variables. This makes it obvious that debugging has the be activated by changing the source code.

Issue #3769, opened 2024-09-11
Use Plack::Middleware::Profiler::NYTProf for profiling the webapp


bschmalhofer:
It appears that the current middleware `$NYTProfMiddleware`, for profiling the web app, is not really functional. Also, the reason why a home grown middleware was created is not really obvious any more. So let's simply use https://metacpan.org/pod/Plack::Middleware::Profiler::NYTProf . The code to activate the middleware can be added as a code comment in _otobo.psgi_. Anybody who tackles profiling can be expected to change some lines of code.

bschmalhofer:
Let's backport to rel-10_1 because it might be useful to compare performance between 10.1 and 11.0.

bschmalhofer:
The primary goal has been achieved. Further improvements can be handled in new issue. Closing this issue now.

bschmalhofer:
Currently profiling the web app with Devel::NYTProf is done by manually editing _bin/psgi-bin/otobo.psgi_ . There the variable `my $ProfilingIsActive = 1;` needs to be set. This is fine for local installations. Especially for autmated testing within GitLab or GitHub it would be nice if profiling could be activated from the outside. The mechanism from #4554 , an additional parameter to _entrypoint.sh_ can be used for that.

This should be targeted for rel-10_1, as that is the first release series that uses the standard middleware for profiling.

PR #3781, opened 2024-09-15
Issue #3769: fix and enhance profiling

bschmalhofer:
There was an error when profiling was deactivated, it propably had to do with the DSL syntax of Plack::Builder.
So let's enable the route /otobo-web/nytprof in all cases.

For convenience create a new report dir for every request. This offers the chance to see all requests, including the AJAX requests. Made a workaround to make the profile id available in the report dirname.

Enhance the profile id to show the human readable date and time.

### DIFF
--- a/bin/psgi-bin/otobo.psgi
+++ b/bin/psgi-bin/otobo.psgi
@@ -100,7 +100,7 @@ eval {
     require Net::DNS;
 };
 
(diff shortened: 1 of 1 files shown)


bschmalhofer:
Ignoring the code policy violation from using `$FindBin`.

Issue #3804, opened 2024-09-22
quick_setup.pl: Include the SystemID in the name of the session cookie


bschmalhofer:
This is sometimes useful during development. Browsing two instances from the same domain is easy when the two instances run on differenent ports and different browsers are used. Using the same browser fails because the session cookies overwrite each other.

A simple solution could be to use unique name for the SessionName and CustomerPanelSessionName SysConfig properties. This could be added to bin/docker/quick_setup.pl . Or is there a better approach?

bschmalhofer:
In this case we want to pass an argument to the quick setup commant. This can be done with a shell function.
```
otobo_docker_quick_setup() {
    docker compose exec web bash -c "date ; hostname ; rm -f Kernel/Config/Files/ZZZAAuto.pm ; bin/docker/quick_setup.pl --db-password otobo_root --http-type http --http-port $1 --system-id $1 --activate-elasticsearch --add-user --add-admin-user --add-customer-user --add-calendar"
}
```

TODO: 

- [x] append the session id to the  session name in the SysConfig
- [x] test wether two installation can run side by side in the same browser
- [x] run the test suite as tests might depend on a specific value for the SystemID
- [x] Define a default value for the SystemID, maybe extract it from the _.env_ file

svenoe:
I understand that the SystemID is in the ticketnumber, precisely for communication between systems, but what would be a reason for having it in the session name?

bschmalhofer:
That is for the case when two Docker based installations run on the same Docker host but on different ports. The two installation can coexist with only one problem. Let's assume that the two installations run on 1481 and 1482.

1. Log into the 1481 installation works fine
2. Log into the 1482 installation in a new tab works fine too
3. Continue work in the 1481 tab breaks as the Cookie has been hijacked by the 1482 tab

bschmalhofer:
The last commits were actually for #3745. I'll rename the issue and mark the old issue as a duplicate. Sorry for causing confusion.

bschmalhofer:
Renaming the session cookie seems to work fine. Using the shell function works as well, at least for my setup. Closing this issue.

PR #3825, opened 2024-09-25
Issue #3804 system

bschmalhofer:


bschmalhofer:
Ignoring the code policy violation because of using `$FindBin`.

PR #3836, opened 2024-09-26
Issue #3559: Added owner set and made lock set depending on sysconfig RequiredLock.

### LINKED ISSUE #3559: AgentTicketQuickClose doesn't lock a ticket
When I close a ticket with QuickClose (AgentTicketQuickClose), in my understanding the ticket should be locked (and therefore an owner should be set) because of the setting Ticket::Frontend::AgentTicketQuickClose###RequiredLock
But that doesn't happen.
This is necessary to save the person who closed the ticket as an owner to the ticket.

I reproduced this on OTOBO 10.1.10 (ov254)

stefanhaerter:
fixes #3559

### DIFF
--- a/Kernel/Modules/AgentTicketQuickClose.pm
+++ b/Kernel/Modules/AgentTicketQuickClose.pm
@@ -135,11 +135,29 @@ sub _SetState {
 
     return if !$Success;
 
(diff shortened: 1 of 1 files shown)


stefanhaerter:
Closed in favor of #3839

PR #3866, opened 2024-10-09
Issue #3719: Providing necessary improvement to previos fix.

### LINKED ISSUE #3719: DynamicField Reference Agent: Empty value doesn't work
When you create a Dynamic Field of type (Reference) Agent the Empty value option doesn't work at all, which causes the alphabetically first agent to be selected automatically.

I reproduced ths on ov270

This is my DF configuration:
image

This is the result (it looks the same e.g. in a Process Activity Dialog or in AgentTicketPhone):
image

stefanhaerter:
Motivated by https://github.com/RotherOSS/ITSMConfigurationManagement/issues/19.

closes #3719


stefanhaerter:
Closing the PR because my opinion is that the root cause needs to be tackled differently.

Issue #3932, opened 2024-11-17
Remove JSON::PP and JSON::PP::Boolan from Kernel/cpan-lib


bschmalhofer:
Wondering whether JSON::PP can be removed from _Kernel/cpan-lib_ .

JSON::PP had been added in OTRS  to _Kernel/cpan-lib_ in 2009. 

```
commit e09bc53d549d57b9d7cfae50e992a797ae67e3b3
Author: Marc Nilius <marc.nilius@otrs.com>
Date:   Thu Dec 31 10:47:25 2009 +0000

    Added cpan module for JSON support, removed JSON function from LayoutAJAX and changed all function calls to the new JSON wrapper.
```
This was fine back than, because JSON::PP was added as a core module only in 2022. See http://perlpunks.de/corelist/mversion?module=JSON%3A%3APP%3A%3ABoolean.

_Kernel/cpan-lib/JSON/PP.pm_ provides the version 2.27203. This is a lower version than the version 2.27300, which is the version supplied in the core of Perl 5.24.0. Perl 5.24.0 is the lowest version of Perl that is supported by OTOBO.

As far as I see, OTOBO does not directly use JSON::PP.  So there is no reason to think that OTOBO needs a specific version of JSON::PP

TODO:

- [x] remove JSON::PP
- [x] remove JSON::PP::Boolean  
- [x] adapt scripts/test/Environment.t where there is an explicit check for JSON::PP

bschmalhofer:
JSON::PP has been removed. Closing this issue.

PR #3933, opened 2024-11-17
Issue #3932: remove PSON::PP from Kernel/cpan-lib

bschmalhofer:


bschmalhofer:
Ran the test suite and found no failures that seem to be related with JSON::PP

PR #2262, opened 2023-04-05
Update Ticket.xml

StefanAbel-OTOBO:
added further description to PostMaster::PreFilterModule###3-NewTicketReject

PR #3956, opened 2024-11-28
Update Ticket.xml

StefanAbel-OTOBO:
Added further description for more clarity


stefanhaerter:
#2262 and #2390 target the same description, thus closing them and only keeping this one.

PR #3962, opened 2024-11-29
Issue #3961: Added parameter to XSLT mapping calls to map undef values and empty strings to empty strings instead of empty hash refs.

### LINKED ISSUE #3961: Change handling of undef values and empty strings in webservice XSLT mapping
The default behavior of XML::Simple is to map undef values or empty strings into empty hash refs. This causes problems e.g. when updating config items via webservice. Decision is to change this, undef values and empty strings should get mapped to empty string.

Further reference: https://metacpan.org/pod/XML::Simple#SuppressEmpty-%3D%3E-1-%7C-%27%27-%7C-undef-%23-in%2Bout-handy

stefanhaerter:
closes #3961

### DIFF
--- a/Kernel/GenericInterface/Mapping/XSLT.pm
+++ b/Kernel/GenericInterface/Mapping/XSLT.pm
@@ -252,6 +252,9 @@ sub Map {
 
     my $XMLSimple = XML::Simple->new;
     my $XMLPre    = eval {
+
+        # Note that the default behavior for SuppressEmpty applies.
+        # This means that attributes with undefined values will be added as empty elements.
         $XMLSimple->XMLout(
             $Param{Data},
             AttrIndent => 1,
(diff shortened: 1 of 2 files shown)


Kernel/GenericInterface/Mapping/XSLT.pm
```diff
@@ -301,10 +302,11 @@ sub Map {
...
             $XMLPost,
-            ForceArray => 0,
-            ContentKey => '-content',
-            NoAttr     => 1,
-            KeyAttr    => [],
+            ForceArray    => 0,
+            ContentKey    => '-content',
+            NoAttr        => 1,
+            KeyAttr       => [],
+            SuppressEmpty => '',
```
bschmalhofer, line 313: Please add a code comment about the intent? If I understand it correctly, that undef values are turned into empty elements not empty attributes.

PR #812, opened 2021-02-19
Implement StartTLS support for LDAP auth modules

LordAro:
Adds ability for LDAP connections to use start_tls, which isn't covered by the Net::LDAP::new function. Only sets/uses the verify parameter, rather than adding all the other parameters, but I'd imagine that's all 99% of people need...

Previous attempts: otrs/otrs#1888 otrs/otrs#1950 Third time lucky?

PR #3989, opened 2024-12-16
Auth::Sync::LDAP was using the wrong config variable

LordAro:
Finally got around to actually removing my original hacks for StartTLS and using what I implemented in #812. Turns out it's slightly broken! Oops.

Fixes Auth::Sync::LDAP config object to look at the correct config object. Others look fine (at least, they match other usages around it)


svenoe:
Hi @LordAro,
thanks for the update, we will have a look and (from my short glance) most likely merge it with the next release.

bschmalhofer (approved):
Yes, the correct config variable should be used here.

Issue #3992, opened 2024-12-17
Install Term::ReadLine::Gnu for history navigation in debugger and Dev::Tools::Shell


bschmalhofer:
When running the Perl debugger, `perl -d`, in an container, then navigation through the command history does not work using the arrow keys. The same applies to the history when running the console command `Dev::Tools::Shell` . According to https://perlmaven.com/debugger-history this can be improved by installing `Term::ReadLine::Gnu` .

I suggest to add this module to the feature `devel:debugging`.

bschmalhofer:
The Docker image for OTOBO 11.1.x is based on Debian slim, which seems to come without readline. So the Debian package _libreadline-dev_ needs to be added as well.

bschmalhofer:
Manual tests of the Perl debugger and of Dev::Tools::Shell looked fine. PR is merged. Closing this issue.

bschmalhofer:
Reopening because of this error message from `perl  bin/otobo.CheckModules.pl --all NoColors `. This was caugth by _scripts/test/CheckModules.t_ .

```
Packages needed for the feature 'devel':
 58   o Data::Dump.......................ok (v1.25)
 59   o Data::Dx.........................ok (v0.000010)
 60   o Term::ReadLine::Gnu..............FAILED! Not all prerequisites for this module correctly installed. 
 61   o Data::Peek.......................ok (v0.52)
 62   o Data::Peek.......................ok (v0.52)
 63   o String::Dump.....................ok (v0.09)
 64   o Selenium::Remote::Driver.........ok (v1.49)
 65   o Test::Compile....................ok (vv3.3.3)
 66   o Test2::Suite.....................ok (v0.000162)
 67   o Test::Simple.....................ok (v1.302199)
 68   o Test2::Tools::HTTP...............ok (v0.12)
 69   o Unicode::GCString................ok (v2013.10)
 70   o Locale::PO.......................ok (v0.27)

```

bschmalhofer:
Merged a fix in https://github.com/RotherOSS/otobo/pull/4012, sorry about misspelling the branch name again, it was not about the issue 2992. Closing this issue again.

bschmalhofer:
Reopening because having a sane debugger is useful on rel-11_0 too.

bschmalhofer:
Cherry-picked the changes into rel-11_0. Tried the Perl debugger in a rel-11_0 container and history navigation worked. Closing this issue again.

PR #3993, opened 2024-12-17
Issue #3993 term readline gnu

bschmalhofer:
This also includes the latest merge from rel-11_0.


bschmalhofer:
This was actually for #3992 .

Issue #4005, opened 2024-12-22
Avoid indirect object syntax in _bin/otobo.ChechModules.pl_


bschmalhofer:
Using the indirect object syntax is discouraged. The case in _bin/otobo.CheckModules.pl_ was find by looking at the git diffs created by the command `find . -name '*.p[lm]' -exec  perl-version-bump --env PERL5LIB=.:Kernel/cpan-lib --version 5.038 {} \;  2>&1 | tee bump.out` .

bschmalhofer:
Tested manually by calling  `bin/otobo.CheckModules.pl --list`. Looks good. PR is merged. Closing this issue.

PR #4006, opened 2024-12-22
Issue #4005: do not use indirect object syntax

bschmalhofer:


bschmalhofer:
Tested manually by calling ` bin/otobo.CheckModules.pl --list`. Looks good.

Issue #4020, opened 2025-01-07
Duplicate content in DynamicFields.xml and OTOBODynamicFields.xml


bschmalhofer:
Before  OTOBO 11.0.1 was released, the config declaration file _Kernel/Config/Files/XML/OTOBODynamicFields.xml_ was renamed to _Kernel/Config/Files/XML/DynamicFields.xml_ . The renaming  was purely for consistency, as the other files in that directory do not have the prefix OTOBO .  See https://github.com/RotherOSS/otobo/commit/d0d02e9711ba5cec8eccd2a9d42e8cb510053b8e . 

After the release of OTOBO 11.0 the renamed file caused a problem when upgrading from 10.1.x to 11.0.y.  During the upgrade process the new files are merged into the previous directory structure. This means that new files are overwritten, but old files are not deleted when they do not exist in the the new directory structure. Thus in the upgraded installation both _DynamicFields.xml_ and _OTOBODynamicFields.xml_ were present and had conflicting content. As a fix the file _OTOBODynamicFields.xml_ war reintruduced in https://github.com/RotherOSS/otobo/commit/119e59fb124928e6daf8731a1922d48ea1f72c02 . Now both files have the same content, which is a bit confusing. The file _OTOBODynamicFields.xml_ is not required to have an settings. So let's turn that file into a dummy, which is also confusing, but in a more consistent way.

There is potential for errors if _OTOBODynamicFields.xml_ has been modified in a OTOBO 10.1.x addon. Avoiding the error is resposibility of whoever ports the OTOBO addon.

See also #663 and #3413 .

bschmalhofer:
Ignoring the complaint from CodePolicy as it feels funny to a add a dummy settings. The test suite looks fine. Closing this issue.

bschmalhofer:
Having an empty OTOBODynamicFields.xml is fine for both native and Docker-based installations. But in the Docker case we can do better. https://doc.otobo.de/manual/installation/11.0/en/content/updating-docker.html says that when upgrading  the command `/opt/otobo_install/entrypoint.sh copy_otobo_next` has to be executed. This command can be enhanced to also remove the useless file _OTOBODynamicFields.xml_ .

bschmalhofer:
Tested the adaption on _entrypoint.sh_ manually. Looks good. PR is merged. Closing this issue again.

bschmalhofer:
Also eliminated that file from the manifest file _ARCHIVE_.

PR #4023, opened 2025-01-07
Issue #4020: Remove settings that are also in DynamicFields.xml

bschmalhofer:
Accept the complaint from CodePolicy about OTOBO::XML::Configuration::XSDValidator


bschmalhofer:
Ignore the complaint from CodePolicy as there are deliberately no settings in _OTOBODynamicFields.xml_ .

Issue #3944, opened 2024-11-20
DynamicField label in front of the value in CustomerUserInterface in DynamicField_Lens


AlexanderOesterling-OTOBO:
### Environment

- Server OS Distribution: 
- OTOBO installation typ: Native
- Browser Chrome
- OTOBO version: 11.0.7

### Expected behavior

I am in CustomerTicketMessage and select a value of a DynamicField_Reference.
I have a DynamicField_Lens that refereneces to a attribute of that selected object in the DynamicField_Reference.

Now the DynamicField_Lens should be filled with the value of that attribute and the label of the DynamicFIeld_Lens should move up so i can see the value.

### Actual behavior

I am in CustomerTicketMessage and select a value of a DynamicField_Reference.
I have a DynamicField_Lens that refereneces to a attribute of that selected object in the DynamicField_Reference.

Now the DynamicField_Lens is filled but the label of the field is in front of the value so i dont see parts of the value.

### How to reproduce

1. Add a DyamicField_Reference and a DynamicField_Lens and link the lens to the reference field.
2. Add them to CustomerTicketMessage.
3. Select an object in the DynamicField_Reference.

### Screenshots

Here is a screenshot that shows my DynamicField_Reference "Sim Card" and also my DynamicField_Lens "Mobil Number".
The label is in front of the actual value of the field. I can only see the last four digits.

![Bildschirmfoto_2024-11-15_14-14-13](https://github.com/user-attachments/assets/1e0518b5-3688-4b70-99e6-e9a394fa5ab1)

svenoe:
I assume the field type of the field the lens points at is "Text", and as for those there was no way to fill them via AJAXUpdate yet, I assume some js is missing which sets the html class. Maybe in the Core.AJAX.js it suffices to simply trigger a `blur` on the field for simple text inputs, where select fields are redrawn etc.

Edit: Textfield is confirmed.

MichaelThumes:
replicated locally on rel-11_0

on initial load of page, the DF lense field renders as:
```HTML
<div class="Field">                    
  <input type="text" class="DynamicFieldText W50pc" id="DynamicField_TicketLense" name="DynamicField_TicketLense" title="TicketLense" value="" aria-invalid="false">
</div>
```

which triggers described erroneous behaviour until the DF is focused by clicking, which changes the rendered HTML to (note the additional css class 'oooFull'!) :
```HTML
<div class="Field oooFull">                    
[...]

PR #4030, opened 2025-01-08
Issue #3944: update CSS class for DF simple text input fields on Ajax…

MichaelThumes:
… Update

closes issue #3944

### DIFF
--- a/var/httpd/htdocs/js/Core.AJAX.js
+++ b/var/httpd/htdocs/js/Core.AJAX.js
@@ -487,6 +487,13 @@ Core.AJAX = (function (TargetNS) {
             if ($Element.hasClass('Modernize')) {
                 $Element.trigger('redraw.InputField');
             }
+            
+            // relevant for customer.pl - trigger a blur after
+            // update to prevent label being displayed in front
+            // of value - see issue #3944
+            if ( $Element.is('input[type=text]')  ) {
+                $Element.trigger('blur');
+            }
         });
     }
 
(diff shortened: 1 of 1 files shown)


stefanhaerter:
Thank you for investigating this - I am quite surprised that `oooFull` seems to be only handled by JavaScript and does not show up anywhere in the templates. Happy to have learned something new ^^

svenoe:
I think the wrong issue is referenced, here (Edit: ah - in the commit 3994, not in the message 3944). I wanted to have a look, because back in my mind, I though we had an issue where something like `$Elemnt.trigger('blur')` was necessary, to trigger something like setting "oooFull". Maybe either way is fine, please quickly consider, whether there might be something missing for other field types, whether oooFull is only set on the parent and not also somewhere on the label or so (just not sure, right now), and probably, if you decide to keep the manual setting of oooFull (which might just be fine), still move it down a bit, at least to after the reference fields, where the element might be of type text, too (also not 100% sure, but definitely text inputs are used in other field types, too - maybe not as the main input which is `$Element`).

MichaelThumes:
alternative solution, which seems to have same effect of adding 'oooFull' css class to parent Element:

at the very end of UpdateFormElements() in Core.AJAX.js:
```Diff
diff --git a/var/httpd/htdocs/js/Core.AJAX.js b/var/httpd/htdocs/js/Core.AJAX.js
index 942080351..42b10dd5c 100644
--- a/var/httpd/htdocs/js/Core.AJAX.js
+++ b/var/httpd/htdocs/js/Core.AJAX.js
@@ -478,14 +478,16 @@ Core.AJAX = (function (TargetNS) {
             // Other form elements
             $Element.val(DataValue);
 
             // Trigger custom redraw event for InputFields
             if ($Element.hasClass('Modernize')) {
                 $Element.trigger('redraw.InputField');
             }
+            
+            $Element.trigger('blur');
         });
     }
```

Hm, maybe this _is_ actually even less invasive?

@svenoe let me know if your prefer this way, I would then update the branch accordingly?

PR #4070, opened 2025-01-15
Issue #4066: Beautified error message for required modules in package manager.

### LINKED ISSUE #4066: Improve error message when a required module is missing during package installation
The current code leads to the following error message if no specific version is required:

Bildschirmfoto_20250114_162608

stefanhaerter:
closes #4066

### DIFF
--- a/Kernel/System/Package.pm
+++ b/Kernel/System/Package.pm
@@ -3908,7 +3908,7 @@ sub _CheckModuleRequired {
                 $Kernel::OM->Get('Kernel::System::Log')->Log(
                     Priority => 'error',
                     Message  => "Sorry, can't install package, because module "
-                        . "$Module->{Content} v$Module->{Version} is required "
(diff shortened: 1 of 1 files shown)


Kernel/System/Package.pm
```diff
@@ -3908,7 +3908,7 @@ sub _CheckModuleRequired {
                 $Kernel::OM->Get('Kernel::System::Log')->Log(
                     Priority => 'error',
                     Message  => "Sorry, can't install package, because module "
-                        . "$Module->{Content} v$Module->{Version} is required "
+                        . "$Module->{Content} " . ( $Module->{Version} ? "v$Module->{Version}" : "" ) . " is required "
```
bschmalhofer: Being a bit picky here. Could you eliminate the extra space when there is no version?

stefanhaerter: Ah, yes, you're right - I thought that I had positioned the quotes to do so, but I overlooked it ;)

stefanhaerter:
Closed in favor of #4084

PR #4084, opened 2025-01-17
Issue #4066: Made version output depending on if a version is given.

### LINKED ISSUE #4066: Improve error message when a required module is missing during package installation

stefanhaerter:
closes #4066


bschmalhofer:
Ignoring complaint from CodePolicy, as it is just about a not installed Perl::Critic module.

Issue #4035, opened 2024-12-27
Update Elasticsearch to image version 7.17.27


MichaelThumes:
[Edit sha]
Original title: ElasticSearch container unhealthy with Kernel version equal or greater 6.12 and systemd equal or greater version 256
[End Edit]

### tl;dr:

Please consider upgrading the otobo-docker Elastic-Search container to latest patchlevel (7.17.26 at time of writing)

### Environment

- Server OS Distribution:Ubuntu 24.10 + Kernel 6.12
- OTOBO installation typ: Docker
- Browser: n/a
- OTOBO version: rel-10_0 through rel -11_1

The crucial point here is the version of the ElasticSearch docker image used, which currently is the same **7.17.3** for all otobo release branches.

### Expected behavior

ElasticSearch container starts up and is reported 'healthy' after initial 'starting health' phase.

### Actual behavior

ElasticSearch container runs into a fatal exception during JVM initialization that kills the JVM and results in a 'docker container unhealthy - restarting' endless loop when run under Kernel >= 6.12 with systemd >= 256.

The issue can be pinned down by looking for a specific fatal exception in the elasticsearch container logs:

docker logs otobo-elastic-1
```
FATAL tini (1)] Unexpected error when forwarding signal: 'Operation not permitted'
2024-12-24 12:10:10,821 main ERROR Could not reconfigure JMX java.lang.NullPointerException: Cannot invoke "jdk.internal.platform.CgroupInfo.getMountPoint()" because "anyController" is null
    at java.base/jdk.internal.platform.cgroupv2.CgroupV2Subsystem.getInstance(CgroupV2Subsystem.java:80)
    at java.base/jdk.internal.platform.CgroupSubsystemFactory.create(CgroupSubsystemFactory.java:114)
    at java.base/jdk.internal.platform.CgroupMetrics.getInstance(CgroupMetrics.java:177)
    at java.base/jdk.internal.platform.SystemMetrics.instance(SystemMetrics.java:29)
    at java.base/jdk.internal.platform.Metrics.systemMetrics(Metrics.java:58)
    at java.base/jdk.internal.platform.Container.metrics(Container.java:43)
    at jdk.management/com.sun.management.internal.OperatingSystemImpl.<init>(OperatingSystemImpl.java:182)
```

This is an error involving the JVM interface to the Kernel cgroups feature. See Additional Information below for details.

### How to reproduce

[...]

PR #4085, opened 2025-01-17
Issue #4035: Updated elasticsearch version tag.

stefanhaerter:
references #4035

### DIFF
--- a/scripts/test/Elasticsearch/TicketCreate.t
+++ b/scripts/test/Elasticsearch/TicketCreate.t
@@ -53,7 +53,7 @@ my $UserID   = 1;
 my $RandomID = $Helper->GetRandomID();
 
 # set necessary sysconfig setting
-$ConfigObject->Set(
+$Helper->ConfigSettingChange(
     Key   => 'CheckEmailAddresses',
     Value => 0,
 );
--- a/scripts/test/Elasticsearch/TicketCreate.t
+++ b/scripts/test/Elasticsearch/TicketCreate.t
@@ -31,11 +31,11 @@ our $Self;
 
 # get needed objects
 my $ConfigObject         = $Kernel::OM->Get('Kernel::Config');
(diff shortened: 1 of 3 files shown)


scripts/test/Elasticsearch/TicketCreate.t
```diff
@@ -0,0 +1,177 @@
...
+use v5.24;
+use utf8;
+
+# core modules
+
+# CPAN modules
+use Test2::V0;
+
+# OTOBO modules
+use Kernel::System::UnitTest::RegisterDriver;    # set up the $Self and $Kernel::OM
```
bschmalhofer: `$Self` is not used in this testscript. So it does not have to be declared and it suffices to load `Kernel::System::UnitTest::RegisterOM`

scripts/test/Elasticsearch/TicketCreate.t
```diff
@@ -51,6 +52,27 @@ my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
 my $UserID   = 1;
 my $RandomID = $Helper->GetRandomID();
 
+# set necessary sysconfig setting
```
bschmalhofer, line 53: I think that the config setting changes should be temporary, using `Kernel::System::UnitTest::Helper::ConfigSettingChange()`.

stefanhaerter: In the meantime, I removed setting 'Elasticsearch::Active', because `Kernel::System::Elasticsearch::InitialSetup()` handles this. I needed to include various other config setting adaptions, which I implemented using `ConfigSettingChange()`.

scripts/test/Elasticsearch/TicketCreate.t
stefanhaerter: Done ✅

Issue #4081, opened 2025-01-16
article css styles with child selectors get broken during handling


chzauleck:
When replacing special characters with HTML Codes in (richtext) articles the css child selector > also gets replaced.

Original:
```
.ck-content .image>figcaption {
    /* stuff */
}
```
In the OTOBO Article:
```
.ck-content .image&gt;

figcaption {
    /* stuff */
}
```

bschmalhofer:
Hi @chzauleck,

I looked into this issue but I think that I do not understand the part about article css styles. I take AgentTicketPhone as an example.  Creating an article with the content

```
.ck-content .image>figcaption {
    /* stuff */
}
```

gives an article that contains

```
.ck-content .image&gt;figcaption {
    /* stuff */
}
```

The `>` is replaced by the HTML entity `&gt`. This substitution is done in the method `Kernel::System::HTMLUtils::Safety()`. This makes sense as we are in text content. There is no tag to be closed by an `>`.

The part that I don't understand is why one would want to add CSS in the message. Internal CSS would be in a `style` attribute where CSS selectors are not supported anyways. Let's discuss that first before I dig deeper into the Safety method.

bschmalhofer:
Discussed this with @chzauleck . The problem with the `>` does not really concern the body of the HTML. This issue is primarily about `<style>` tags in the HTML header. Replacing an `>` with `&gt;` does break the declared CSS.

The cause of the effect seems to be in the `Safety()` method. More precisely in the underlying [HTML::Scrubber](https://metacpan.org/pod/HTML::Scrubber) module. See https://metacpan.org/dist/HTML-Scrubber/source/lib/HTML/Scrubber.pm#L300 . See also the discussion in https://github.com/nigelm/html-scrubber/issues/1 , https://github.com/nigelm/html-scrubber/issues/18 and in https://rt.cpan.org/Public/Bug/Display.html?id=2991 .

Incidently OTOBO already uses a patched version of HTML::Scrubber. So special casing that case should be feasible.

PR #4102, opened 2025-01-22
Issue #4081 html scrubber

bschmalhofer:


bschmalhofer:
Closing, as using a state variable is a bad idea when multiple instances of HTML::Scrubber could be active at the same time.

Issue #4106, opened 2025-01-27
Remove Net::SSLGlue::POP3 and Net::SSLGlue::SMTP from Kernel/cpan-lib


bschmalhofer:
The `Net::SSLGlue::*`modules kick in when very old `Net::*` modules do not provide a `starttls()` method.  The minimum version of Perl that is supported by OTOBO is Perl 5.24. Perl 5.24 provides libnet 3.08 as a core module. The modules Net::POP3 and Net::SMTP are provided by libnet and they have a `starttls()` method. Therefore the `Net::SSLGlue` stuff is no longer needed. 

Removing these modules in _Kernel/cpan-lib_ poses no problem under Docker with lingering files. _Kernel/cpan-lib_ is rebuilt when OTOBO is upgraded.

bschmalhofer:
The test suite looks fine. PR is merged. Closing this issue.

bschmalhofer:
Removing the Net::SSLGlue modules can already be done in rel-10_1. The goal is to reduce the differences between the three current release series.

bschmalhofer:
The test suite still looks fine. PR is merged und upmerged. Closing this issue again.

bschmalhofer:
This change will be reverted. The interface should not be changed in patch level releases.

bschmalhofer:
The changes have been reverted. Closing this issue again.

bschmalhofer:
Reopening and changing the milestone to 11.1.1.  Net::SSLGlue should only be removed for the rel-11_1 release series. Again this is because we don't wont interface changes in patch level revisions.

bschmalhofer:
The Net::SSL modules, and the releated Code, are now only removed in rel-11_1. 

Removing cruft from _Kernel/cpan-lib_ in rel-11_1 is fine, as there is no release yet in the 11.1. release series.

Closing this issue.

PR #4109, opened 2025-01-27
Issue #4106 net sslglue

bschmalhofer:


bschmalhofer:
Ignoring Code::Policy violation for IO::Socket::SSL. This is a module from CPAN, not sure why it is checked.

PR #4125, opened 2025-02-07
Issue #4124: Improved POD of VirtualFS Read sub.

### LINKED ISSUE #4124: Clarify POD of VirtualFS Read sub
The POD is technically correct, however very brief about the content value it returns.

stefanhaerter:
closes #4124

### DIFF
--- a/Kernel/System/VirtualFS.pm
+++ b/Kernel/System/VirtualFS.pm
@@ -78,7 +78,7 @@ read a file from virtual file system
 
 returns
 
-    Note that the content value is returned as a reference to a string.
+Note that the content value is returned as a reference to a string.
 
     my %File = (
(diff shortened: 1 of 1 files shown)


Kernel/System/VirtualFS.pm
```diff
@@ -78,8 +78,10 @@ read a file from virtual file system
 
 returns
 
+    Note that the content value is returned as a reference to a string.
```
bschmalhofer: That comment line is indented. Therefore it is rendered as code. So it should be preceded by a '#'. Or remove the indention for having that comment as regular POD text.

stefanhaerter: Done 👍🏻

PR #4168, opened 2025-02-19
Refactored BaseEntity.pm

rmfalves:


Kernel/System/DynamicField/Driver/BaseEntity.pm
```diff
@@ -678,12 +680,15 @@ sub ReadableValueRender {
...
-
+    my $ItemList = $Self->PossibleValuesGet(
+        %Param
+    );
     my @ReadableValues;
-
     for my $Item (@Values) {
         $Item //= '';
-
+        if ($Item) {
```
bschmalhofer, line 689: Is the hard coded key 'Item' really wanted here? 
It would also help me to have a more descriptive name of `$ItemList`. It looks like a mapping but it is not obvious to me what is mapped to what.

stefanhaerter:
Discussed with @rmfalves. Conclusion is that this PR can be closed.

PR #4242, opened 2025-03-27
Issue #4241: Removed setting Used checkbox again.

### LINKED ISSUE #4241: Date and DateTime dynamic fields do not respect default value when shown via ACL
When dynamic fields of type date and date time are hidden and then shown via ACL, the checkbox is always set and does not respect the default value.

Related: #3483

stefanhaerter:
closes #4241


stefanhaerter:
Closed in favor of #4251

Issue #4174, opened 2025-02-21
Eliminate usage Net::IMAP::Simple, only use Mail::IMAPClient for IMAP MailAccount


bschmalhofer:
The modules `Kernel::System::MailAccount::POP3`,  `Kernel::System::MailAccount::POP3S`, and `Kernel::System::MailAccount::POP3TLS` have a nice architecture. They are all based on `Ǹet::POP3` and POP3S and POP3TLS inherit from POP3. 

Things are not so nice for IMAP. `Kernel::System::MailAccount::IMAP` and `Kernel::System::MailAccount::IMAPS` are both based on `Net::IMAP::Simple` and IMAPS inherits from IMAP. But `Kernel::System::MailAccount::IMAPTLS` is an outlier. This module uses `Mail::IMAPClient`. I think that there is at least one CPAN module that supports both IMAPS and IMAP+STARTTLS. So let's use a similar approach as for POP3 and eliminate some code. 

TODO:

- [x] find best suited module for IMAP, IMAPS, and IMAP+STARTTLS, Net::IMAP::Client is another option
- [x] refactor the IMAP[|S|TLS] modules
- [ ] optionally add tests that actually fetch mails from a mocked mail server

bschmalhofer:
Closing as the issue is on the wishlist.

bschmalhofer:
I have a  preference for using Mail::IMAPClient as this module seem to receive the most care and because the IDLE command is supported there. Who would want to wait for emails?

bschmalhofer:
Let's do this for rel-11_1.

The first PR will just be a preparation for the actual changes. The plan is:

- [x] _Kernel/System/MailAccount/IMAP.pm_ will basically be a copy of _IMAPTLS.pm_ with neither the *Starttls* nor the *Ssl*  option
- [x] _IMAPTLS.pm_ will inherit from the changed IMAP.pm, with the *Startlls* option passed
- [x] _IMAPS.pm_ will  inherit from the changed IMAP.pm, with the *Ssl* option passed

bschmalhofer:
Commits show up multiple times. Propably because I renamed the branch locally and pushed it again.

bschmalhofer:
Reopen as this should be in OTOBO 11.1

bschmalhofer:
A preliminary implementation is complete. But still needs to be tested.

bschmalhofer:
[...]

PR #4248, opened 2025-03-31
Issue #4174 mailaccount step1

bschmalhofer:


bschmalhofer:
The CodePolicy failure is about something in IO::Socket::SSL, see https://github.com/noxxi/p5-io-socket-ssl/issues/155. I don't think that we have to be concerned about this.

PR #4394, opened 2025-06-11
Issue #4393: prevent wrong loginform from being displayed 

### LINKED ISSUE #4393: OpenIDConnect wrong Login Form and misleading error msg on browser restart
### Environment

- Server OS Distribution: all
- OTOBO installation typ: docker with OpenIDConnect - keycloak
- Browser: mostly chrome
- OTOBO version: 10_1 upwards

### Expected behavior

after logout and restarting the browser user is shown the SSO login of the OIDC provider when auto-reopening browser tabs

### Actual behavior

user is shown the classic otobo login form with a misleading error msg. very bad UX.

### How to reproduce

replicated in Chrome with "auto re-open browser tabs" setting enabled
OIDC provider is keycloak

1. Log into Otobo using keycloak 
3. Open Otobo pages in several tabs
4. Log out in Otobo, and confirm on the keycloak "do you want to log out" page
5. Close Browser without closing tabs first
6. Restart Browser
7. Login in keycloak
8. Note that the other tabs now re-load but run into an error condition because of invalid OIDC state, showing the not very helpful otobo standard login form with a not helpful error message, confusing customers

### Additional information

Behaviour seems to be slightly different between Browsers and OS. Under Linux only the other open tabs ran into the error condition, under MacOS we have seen it for all tabs of the user.

MichaelThumes:
PR for bugfix in rel-10_1. Do **not** upmerge, dedicated PRs will be made for the higher release branches once this one has completed code-review.

- prevent display of wrong login form _if_ OIDC is enabled
- support multiple tabs for the OIDCCSRF cookie

closes #4393

### DIFF
--- a/Kernel/System/Web/InterfaceAgent.pm
+++ b/Kernel/System/Web/InterfaceAgent.pm
@@ -326,24 +326,13 @@ sub Content {    ## no critic qw(Subroutines::RequireFinalReturn)
         # login is invalid
         if ( !$User ) {
 
-            my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
-
-            if ( ref $AuthObject->{AuthBackend} eq 'Kernel::System::Auth::OpenIDConnect' ) {
-
-                # throw a Kernel::System::Web::Exception that redirects
-                $LayoutObject->Redirect(
-                    OP    => "",
-                    Login => 1,
-                );
-                return;
-            }
-
             my $Expires = '+' . $ConfigObject->Get('SessionMaxTime') . 's';
             if ( !$ConfigObject->Get('SessionUseCookieAfterBrowserClose') ) {
                 $Expires = '';
             }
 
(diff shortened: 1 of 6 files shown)


MichaelThumes:
testsuite run completed green on the working branch for rel-10_1

Kernel/System/Web/InterfaceAgent.pm
```diff
@@ -616,6 +627,14 @@ sub Content {    ## no critic qw(Subroutines::RequireFinalReturn)
                 );    # throws a Kernel::System::Web::Exception
             }
 
+            # try auth module specific logout
+            my $LogoutInfo = $Kernel::OM->Get('Kernel::System::Auth')->Logout();
```
bschmalhofer: Would be nice if that method were called `GetLogoutInfo()`, but that ship has sailed.

bschmalhofer:
I only took a superficial look. Looks good to me. I like that there is more red than green.

Issue #4360, opened 2025-05-22
Set ticket values for (re)appearing fields in AgentTicketActionCommon and related masks.


svenoe:
When a dynamic field is not rendered directly but only shown in the HideShow process, for masks working on an existing ticket (ATAC, ATC,...), we should render the value of the ticket and not have it pop up empty.

In the same vein please check the behavior of initial rendering and ajax updates (as well as storing) for dynamic fields which have content restricted by another attribute (DF) filled in the ticket, but not shown on the mask. At least the initial rendering and the ajax update behave different.

Edit: Fix for 11.0 only, as we did some changes in the logic concerning standard fields already and new systems are likely using 11.0 anyways. (And if need be, we can backport.)

stefanhaerter:
The possible fix, if accepted, needs to be implemented in `Kernel/System/ITSMConfigItem/FieldRestrictions.pm` in the [ITSMConfigurationManagement](https://github.com/RotherOSS/ITSMConfigurationManagement) package as well.

stefanhaerter:
> When a dynamic field is not rendered directly but only shown in the HideShow process, for masks working on an existing ticket (ATAC, ATC,...), we should render the value of the ticket and not have it pop up empty.

This part is achieved with commit https://github.com/RotherOSS/otobo/commit/652a5884dbc74c0546e1c0fdcaca085f6cfa96ba.

> In the same vein please check the behavior of initial rendering and ajax updates (as well as storing) for dynamic fields which have content restricted by another attribute (DF) filled in the ticket, but not shown on the mask. At least the initial rendering and the ajax update behave different.

I was not yet able to reproduce this. It is indeed the case that the stored value for a hidden dynamic field is not taken into account if 'Properties' is used in the ACL, but it works if 'PropertiesDatabase' is used.

I will proceed in testing this.

stefanhaerter:
Separated the first change in https://github.com/RotherOSS/otobo/pull/4397.

I did so because I consider the second change necessary to be a considerable change in behavior.

stefanhaerter:
PR merged, closing the issue.

PR #4397, opened 2025-06-12
Issue #4360: Prioritize existing dynamic field values over default values when un-hiding dynamic fields.

stefanhaerter:
references #4360

### DIFF
--- a/Kernel/System/Ticket/FieldRestrictions.pm
+++ b/Kernel/System/Ticket/FieldRestrictions.pm
@@ -293,6 +293,9 @@ sub GetFieldStates {
                     if ( !$ValueIsDifferent ) {
                         $UpdateRequired = 0;
                     }
+                    else {
+                        $UpdateRequired = 1;
+                    }
                 }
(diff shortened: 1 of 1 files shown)


Kernel/System/Ticket/FieldRestrictions.pm
```diff
@@ -271,12 +273,40 @@ sub GetFieldStates {
...
+
+                if ( defined $TicketData{"DynamicField_$DFName"} ) {
+
+                    my $ValueIsDifferent = $Kernel::OM->Get('Kernel::System::DynamicField::Backend')->ValueIsDifferent(
+                        DynamicFieldConfig => $DynamicFieldConfig,
+                        Value1             => $DFParam->{"DynamicField_$DFName"},
+                        Value2             => $TicketData{"DynamicField_$DFName"},
+                    );
+
+                    if ( !$ValueIsDifferent ) {
```
svenoe, line 293: We need if and else

Issue #4399, opened 2025-06-12
Avoid useless warning in AdminAppointmentCalendarManage frontend


bschmalhofer:
A check for the existence of the GroupID of an calendar had been added in an OTOBO package. This checks seems to be sensible in the OTOBO core as well.

bschmalhofer:
PR is applied. No further tests done. Closing anyways.

PR #4400, opened 2025-06-12
Issue #4399 avoid warning

bschmalhofer:


bschmalhofer:
Ignoring the CodePolicy error, just some dependency that could not be installed

PR #4404, opened 2025-06-13
Issue #2539: Made sysconfig OpenMainMenuOnHover available for user modification.

### LINKED ISSUE #2539: Shift OpenMainMenuOnHover to UserPreferences
It is great to have OpenMainMenuOnHover as a sysconfig option, but I would appreciate it even more to have it as UserPreference. I would suggest to keep the sysconfig option as default, but to add a UserPreference which can overrule it.

stefanhaerter:
closes #2539


bschmalhofer:
Merging as the change has currently no user visible effect.

PR #4414, opened 2025-06-18
Issue #4413: Used correct variable for check.

### LINKED ISSUE #4413: AdminProcessManagement: Wrong variable used in check
There is a confusion between `$EntityID` and `$StateEntityID` in the AdminProcessManagement module.

stefanhaerter:
closes #4413

### DIFF
--- a/Kernel/Modules/AdminProcessManagement.pm
+++ b/Kernel/Modules/AdminProcessManagement.pm
@@ -756,7 +756,7 @@ sub Run {
         my $StateEntityID = $StateLookup{'Inactive'};
 
         # show error if  StateEntityID for Inactive does not exist
-        if ( !$EntityID ) {
+        if ( !$StateEntityID ) {
             return $LayoutObject->ErrorScreen(
                 Message => Translatable('The StateEntityID for state Inactive does not exists'),
             );


Kernel/Modules/AdminProcessManagement.pm
```diff
@@ -756,7 +756,7 @@ sub Run {
         my $StateEntityID = $StateLookup{'Inactive'};
 
         # show error if  StateEntityID for Inactive does not exist
-        if ( !$EntityID ) {
+        if ( !$StateEntityID ) {
```
bschmalhofer, line 759: Are you sure that this does not change existing behavior? I looked into the table _ticket_state_ and did not see the state name 'Inactive'.

stefanhaerter: The state stuff in this place does not refer to Kernel::System::State, but to Kernel::System::ProcessManagement::DB::Process::State - a bit confusing, I know...

bschmalhofer: Talked with @stefanhaerter about this. The `$StateObject` is not Kernel::System::State but Kernel::System::ProcessManagement::DB::Process::State which does have the hard coded state 'Inactive'.

PR #4417, opened 2025-06-18
Issue #4416: Provided initial value for variable to prevent warning.

### LINKED ISSUE #4416: Provide initial value for variable in AdminSystemConfiguration module to prevent warning
```
Use of uninitialized value $ExtraParams in concatenation (.) or string at /opt/otobo/bin/psgi-bin/../../Kernel/Modules/AdminSystemConfiguration.pm line 601.
```

stefanhaerter:
closes #4416

### DIFF
--- a/Kernel/Modules/AdminSystemConfiguration.pm
+++ b/Kernel/Modules/AdminSystemConfiguration.pm
@@ -583,7 +583,7 @@ sub Run {
             UserID            => $Self->{UserID},
         );
 
-        my $ExtraParams;
+        my $ExtraParams = '';
 
         if ( !$ConfigurationLoad ) {
 


Kernel/Modules/AdminSystemConfiguration.pm
```diff
@@ -583,7 +583,7 @@ sub Run {
             UserID            => $Self->{UserID},
         );
 
-        my $ExtraParams;
+        my $ExtraParams = '';
```
bschmalhofer, line 586: Looks sane. It is a bit unusual that the URL part  passed to `Redirect()` in the lines below has a trailing semicolon. But that has nothing to do with the warning.

Issue #4351, opened 2025-05-20
Error when importing appointments (Internal Server Error).


KhalilHamada-OTOBO:
OTOBO Version: 11.0.x.

When importing calendar appointments (.ics files) into OTOBO, an "Internal Server Error" message appears.
However, the calendar entries are imported correctly despite the error message.
This issue only affects version 11.0.x, version 10.1.x is not affected. For further information, please feel free to contact me.

stefanhaerter:
Is there any log message present in the system logs when this happens? If so, can you please post it here?

stefanhaerter:
Log entry:

```
ERROR: OTOBO-CGI-10 Perl: 5.38.4 OS: linux Time: Tue May 20 07:06:28 2025

 Message: Invalid value for TimeZone: Europe/Amsterdam.

 RemoteAddress: 192.168.65.1
 RequestURI: /otobo/index.pl

 Traceback (19): 
   Module: Kernel::System::DateTime::_CPANDateTimeObjectCreate Line: 2166
```

After a short investigation, timezone 'Europe/Amsterdam' is indeed missing from the list of timezones. Found #2149, but according to the issue, the module should not have been updated.

stefanhaerter:
The actual problem stems from the timezone 'Europe/Amsterdam' being removed from the list of timezones and thus the host system not providing it in the list of timezones to validate against.

Suggestion by @bschmalhofer was to implement a mapping from 'Europe/Amsterdam' to 'Europe/Brussels' during the import, which is a good solution in my opinion.

stefanhaerter:
As the issue stems from the host system, all OTOBO versions may be affected, thus setting the milestone to 10.0.

PR #4441, opened 2025-07-02
Issue #4351: Fixed deprecated timezone ids

rmfalves:


bschmalhofer:
This fix is only concerned about the Import from iCal files. So there is no risk that other things break. This is good. Merging this issue.

Issue #4476, opened 2025-07-14
Fix spelling: appoitnment => appointment


bschmalhofer:
Found a speling eror:

```
grep -r appoitnment Kernel/ | grep -v Kernel/Language
grep: Kernel/Output/HTML/Templates/Standard/.AdminAppointmentNotificationEvent.tt.swp: Übereinstimmungen in Binärdatei
Kernel/Output/HTML/Templates/Standard/AdminAppointmentNotificationEvent.tt:                <label for="IncludeInvalid">[% Translate("Include invalid appoitnment notifications") | html %]</label>

```

bschmalhofer:
PR is merged. Closing this issue.

PR #4477, opened 2025-07-14
Issue #4476: fix spelling appoitnment => appointment

bschmalhofer:
the correct spelling was already in the translation files. So only the invalid string was removed from the translation files.


bschmalhofer:
Ignoring error in CodePolicy

PR #4509, opened 2025-07-22
Issue #4067: added mechanism to style iframe content for customer ui

### LINKED ISSUE #4067: Font family in iframes (AGB text and RichText tile in dashboard) are now Quicksand
Seen here for example:

fontfamily

First assumption was that the tile content gets inserted with editor css upon installation, but this doesn't hold true:

https://github.com/RotherOSS/otobo/blob/dfbc2e6c3a42af6161c256557e335eec6b202e94/scripts/database/otobo-initial_insert.xml#L4324-L4331

chzauleck:
that means quicksand font for things like the about or the info tile on the customer dashboard


bschmalhofer (commented):
Looks good to me

Issue #4528, opened 2025-07-28
Check in the test suite wheter strictures and warnings are enabled


bschmalhofer:
> Got into a rabbit hole investigating why that "Possible precedence problem" did not show up in _scripts/test/Compile.t_ . Turns out that the verbose mode of `Test::Compile::Internal` must be active for compile time warnings to be printed as diagnostics. Also `use warnings;` must be active  in order to the warning being emitted.
> 
> The other thing is that the test suite does not check whether strictures and warnings are enabled. The CodePolicy does check that. But I think that this is important enough to warrant some extra CPU cycles and disk reads in the test suite. 

 _Originally posted by @bschmalhofer in [#4452](https://github.com/RotherOSS/otobo/issues/4452#issuecomment-3126816963)_

Those checks can be done with `Test::Strict`which would be a new _devel:test_ dependency.

bschmalhofer:
Things look fine. PR is merged. Closing this issue.

bschmalhofer:
A bit related is [Test::Warnings](https://metacpan.org/pod/Test::Warnings). This testing module checks whether the test script, including the loaded modules, emit any warnings. It effectively adds another test case to the test script. So it would be useful to add `use Test::Warnings;` to test scripts. Maybe even to _scripts/test/Compile.t_  .

 Test::Warnings is already a dependence in the devel:test feature as Data::Peek requires it. But let's make this dependency explioit in _bin/otobo.CheckModules.pl_ .

PR #4529, opened 2025-07-28
Issue #4528 test strict

bschmalhofer:


bschmalhofer:
Ignoring complaints by CodePolicy about the usage of FindBin.

PR #4608, opened 2025-08-22
Issue #4596: Delete meta article if inserting into article_data_mime table fails.

### LINKED ISSUE #4596: Tidy article table if article creation fails mid way.
If article creation fails because of e.g. #4595 we should not simply return in https://github.com/RotherOSS/otobo/blob/8cdaafa1693a83ae6b217805ebc81daa005c9311/Kernel/System/Ticket/Article/Backend/MIMEBase.pm#L812 or the select below, but have an additional method `_MetaArticleDelete()` to clean up the `article` table.

In case of an incoming mail, the mail will be stored in `var/spool/` anyways, having the incomplete article in the ticket will result in error messages every time the ticket is accessed, and there is no good way to tidy this for the users.

stefanhaerter:
closes #4596


bschmalhofer (approved):
Looks good to me.

I checked the return value of DBI::do(). In case no insert takes place `0E0` is returned which evaluates to true in boolean context. This is fine.

stefanhaerter:
Thanks a lot for reviewing and checking, @bschmalhofer =)

Issue #4619, opened 2025-08-25
Sanitize input for backup.pl. Thanks to Diego Berger Tellaroli


MichaelThumes:
sanitize all user input to backup.pl

MichaelThumes:
merged

Issue #4570, opened 2025-08-06
Early exit when the backup directory is within /opt/otobo


bschmalhofer:
When running _scripts/backup.pl_ without the option `-d` :

```
$ ./scripts/backup.pl 
Backup /opt/otobo/2025-08-06_15-29-34/Config.tar.gz ... done
Backup /opt/otobo/2025-08-06_15-29-34/Application.tar.gz ... tar: ./2025-08-06_15-29-34/Application.tar.gz: file changed as we read it
tar: ./.gnupg/S.gpg-agent.extra: socket ignored
tar: ./.gnupg/S.gpg-agent.browser: socket ignored
tar: ./.gnupg/S.gpg-agent: socket ignored
tar: ./.gnupg/S.gpg-agent.ssh: socket ignored
failed
Deleting incomplete backup /opt/otobo/2025-08-06_15-29-34 ... done
Backup failed at ./scripts/backup.pl line 266.


```

It would be nice of there was a nice message stating that the directory option is required and that it should be a directory that is part of the backup.

bschmalhofer:
Set the milestone to 10.0 as it is not urgent, but still relevant for all release series.

MichaelThumes:
merged

PR #4621, opened 2025-08-25
Issue #4619: sanitize user input to backup.pl

MichaelThumes:
closes https://github.com/RotherOSS/otobo/issues/4619
closes https://github.com/RotherOSS/otobo/issues/4570

sanitize user input to backup.pl script.

-    validate input using regex for allowed characters
-    quote variables in generated system command where appropriate
-    check backup directory is not under $OTOBO_HOME

### DIFF
--- a/scripts/backup.pl
+++ b/scripts/backup.pl
@@ -21,18 +21,18 @@ use v5.24;
 use utf8;
 
 # use ../ and ../Kernel/cpan-lib as lib location
-use FindBin qw($RealBin);
+use FindBin qw($RealBin);    ## no perlimports, not sure why perlimports wants $Dir
 use lib "$RealBin/..";
 use lib "$RealBin/../Kernel/cpan-lib";
 
 # core modules
 use Getopt::Long qw(GetOptions);
-use Cwd qw(getcwd abs_path);
+use Cwd          qw(abs_path getcwd);
 
 # CPAN modules
 
 # OTOBO modules
-use Kernel::System::ObjectManager;
+use Kernel::System::ObjectManager ();
 
 # file scoped option variables
 my (
@@ -47,7 +47,7 @@ my (
     $ExtraDumpOptions,
 );
 my $MaxAllowedPacket = '64M';          # 64 Megabytes is fine as the default, as that is already required on the server side
-my $BackupDir        = getcwd();
+my $BackupDir;
 my $BackupType       = 'fullbackup';
 
 sub Main {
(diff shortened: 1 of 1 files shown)


scripts/backup.pl
```diff
@@ -219,14 +219,24 @@ sub Main {
 }
 
 # create new backup directory
+
+$BackupDir = abs_path($BackupDir);
+
 my $Home = $Kernel::OM->Get('Kernel::Config')->Get('Home');
 
+if( $BackupDir =~ /^$Home/ ) {
```
bschmalhofer, line 227: Here I would do a more general check using https://metacpan.org/pod/Path::Class::Dir#$boolean-=-$dir-%3Econtains($other). But this can only be added in rel-10_1 as `Path::Class` is only required since OTOBO 10.1.1.

scripts/backup.pl
```diff
@@ -363,11 +373,11 @@ sub Main {
         }
 
         if ($DatabaseHost) {
-            $DatabaseHost = "-h $DatabaseHost";
```
bschmalhofer, line 366: Additionally to restricting the hostname and the other variables, I think  that using `system` in list form would be an useful approach here. See https://www.perl.com/article/quoting-the-shell/ .

bschmalhofer (commented):
Could you split up the PR in different commits? In order to differentiate between the tidyings and the functional changes.

scripts/backup.pl
```diff
@@ -47,7 +47,7 @@
     $ExtraDumpOptions,
 );
 my $MaxAllowedPacket = '64M';          # 64 Megabytes is fine as the default, as that is already required on the server side
-my $BackupDir        = getcwd();
+my $BackupDir;
```
bschmalhofer, line 50: Is removing the default for the backup dir intended?

MichaelThumes: given that docker default dir is /opt/otobo, I found the default annoying. but not everything is dockerized, so I am happy to put it back

scripts/backup.pl
MichaelThumes: system in list form is great for _single commands_, but would it work for complex commands containing pipes and redirections ?

scripts/backup.pl
MichaelThumes: happy to craft a separate PR for 10_1 upward, yes.

MichaelThumes:
closing this PR after discussion and will split into separate ones

scripts/backup.pl
bschmalhofer: Yes, I guess the a plain `system` is no good for complex commands. I recall that I once used `IPC::Run` or such to do the more complex stuff. But that would be overkill here.

scripts/backup.pl
bschmalhofer: Yes, I'm only wary of changing the behavior in a patch level release. After all `backup.pl` can also be called from the backup directory itself.

Issue #4606, opened 2025-08-21
Links on dynamic fields do not work on Set-inner dynamic fields


stefanhaerter:
Observed on CIs and Tickets:

- DynamicField Text, Link configured with Template::Toolkit expression (e.g. `[% Data.FieldName %]`) - pre- or absence of namespace is irrelevant
- Place Text-DF in DynamicField Set (doesn't matter if multi- or single-value)
- Create a ticket or CI with an URL filled in the Text-DF
- View the Item in the Zoom View
- Link is not being interpolated

MichaelThumes:
first PR https://github.com/RotherOSS/otobo/pull/4648

svenoe:
This is a bit more complicated, unfortunately. Currently DisplayValueRender is agnostic in terms of ticket data etc.. The link is not rendered by the field driver, but rather by the calling method (i.e. TicketInformation, AgentConfigItemZoom,... there are many).

Currently the "consistent" option I see, would be, to just enable the usage of other field information within the set index, but that won't cover most relevant use cases, I assume. In my opinion this is a topic we should postpone and pick up again, if/when we change how links in DisplayValueRender are rendered also for MultiValue-Fields (especially reference fields), which also is a topic, and which also needs a more fundamental refactoring of how this method is used.

We will discuss this internally.

stefanhaerter:
Decision was to implement this for 11.1 as this requires structural changes in EditFieldRender mechanisms.

PR #4648, opened 2025-08-28
Issue #4606: interpolate links in set-inner DFs

MichaelThumes:
first attempt to resolve this issue.

Basically, since $Element->{Link} might contain Template Toolkit expression(s), we need to interpolate before construction the HTML. so far so good.

remaining issue is the fact that for this to succeed in all scenarios the help text for advertises in the DF UI, we need more data as the link cannot only reference the DF itself, but also other DFs in the Ticket containing the DF.

For the TicketInformation Widget int AgentTicketZoom, this can be solved by passing the Ticket data along in the call to DisplayValueRender(), and that seems to work nicely.

@svenoe  
Worried a bit if there are other known places besides AgentTicketZoom where passing additional data would also be necessary for the full feature set?


svenoe:
I would like to postpone this. See my comment in the issue.

PR #4657, opened 2025-09-01
Issue #4130: avoid setTimeout when calling InitFavourites() after AJAX call

### LINKED ISSUE #4130: System Configuration: saving a setting as a favorite doesn't work
When you tray to save any setting (I tested it with CustomerColorDefinitions) as a favorite, it jumps a bit upwards and doesn't save it.

1. Go to Admin > System Configuration
2. chose any setting of your liking
3. click the setting name for the "dropdown" to appear
4. click "Add as favorite" and see that it jumps up, and doesn't actually save it as a favorite.

So I cannot actually save the setting as a favorite at all.

Tested on Demo system with OTOBO 11.0.6 and on customer's system 11.0.7 with macOS Safari and Windows Microft Edge, also after Maint::Cache::Delete and Maint::Loader::CacheCleanup.

MichaelThumes:
closes #4130

### DIFF
--- a/var/httpd/htdocs/js/Core.SystemConfiguration.js
+++ b/var/httpd/htdocs/js/Core.SystemConfiguration.js
@@ -1714,12 +1714,11 @@ var Core = Core || {};
                         Counter++;
                     },
                 });
+
+                Core.Agent.Admin.SystemConfiguration.InitFavourites();
+
             }, 'html'
         );
-
-        window.setTimeout(function() {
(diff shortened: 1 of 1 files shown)


chzauleck (approved):
makes me suspicious why it wasn't put in like that in the first place instead of the weird timeout thing, but i can't find any reason so let's go for it

rmfalves (approved):
I agree, this looks like a cleaner solution than the Timeout approach (maybe the timeout was some kind of workaround found for some problem with events, but that is just a guess).

Issue #4415, opened 2025-06-18
Process Import: EntitySyncState is not set


stefanhaerter:
The subactions ProcessCopy, ProcessNewAction, ProcessEditAction and ProcessDelete do set the EntitySyncState, but ProcessImport doesn't do so. Instead, it 'fakes' the mechanism by setting a notification, but the notification does disappear on re-navigating to the page.

stefanhaerter:
Additional finding: This also affects importing Ready2Adopt processes.

rmfalves:
Solved, but along with the info in the DB, the "fake" sync message remains in the code so that it is displayed as soon as the import is finished. Otherwise, it becomes harder to combine both messages when one comes via notifications and the other from the DB. On subsequent navigations to the page, however, the DB inspection mechanism takes over and the sync message is displayed.

(And thanks @MichaelThumes for guiding me through some basic concepts of processes.)

PR #4662, opened 2025-09-02
Issue #4415: Process Import: EntitySyncState is not set

rmfalves:


MichaelThumes (approved):
Looks good to me.

Due to my limited knowledge in this field, setting @svenoe as a second reviewer

svenoe:
Can we just set the SyncState directly in `Kernel/System/ProcessManagement/DB/Process.pm` after ProcessUpdate() directly? Firstly we would not have to change the return structure, additionally there might be a Console script upcoming for process imports, also, and we would need that in this case, too.

(Additionally, more out of curiosity, if I set up two processes, which use the same ActivityDialog, e.g.. After saving and synchronising everything, what happens if I change this ActivityDialog in one process and save it. Are both processes set to not synchronized? I guess it does not really matter, in the end, though; if I synchronise them via the frontend, both should have the altered Dialog, anyways, don't they?)

Changed the decision here: When importing via console script we will roll out and synchronise the processes directly, anyways.

PR #4663, opened 2025-09-03
Issue 4542 unittests for ticket mask

MichaelThumes:
first set of unittests for Ticket::Mask, handling the current checks in Ticket::Mask

there is a lot more potential for further tests, but Ticket::Mask probably should get more sanity checks first.

### DIFF
--- a/scripts/test/Ticket/TicketMasks.t
+++ b/scripts/test/Ticket/TicketMasks.t
@@ -1,0 +1,295 @@
+# --
+# OTOBO is a web-based ticketing system for service organisations.
+# --
+# Copyright (C) 2019-2025 Rother OSS GmbH, https://otobo.io/
+# --
+# This program is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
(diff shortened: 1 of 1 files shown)


scripts/test/Ticket/TicketMasks.t
```diff
@@ -0,0 +1,300 @@
...
+# --
+
+use strict;
+use warnings;
+use utf8;
+
+use Test2::V0;
+use Kernel::System::VariableCheck qw(:all);
+
+# Set up the test driver $Self when we are running as a standalone script.
```
bschmalhofer: I recommend to use `Kernel::System::UnitTest::RegisterOM` and do away with the almost unused `$Self`.

scripts/test/Ticket/TicketMasks.t
```diff
@@ -0,0 +1,300 @@
...
+
+            ok( IsHashRefWithData($Definition), "Got Definition" );
+
+            my $Count = scalar keys $Definition->{DynamicFields}->%*;
+            ok( $Count == $TestCase->{DFCount}, $TestCase->{DFCount} . ' DFs have beend expected, got: ' . $Count );
+        }
+    };
+}
+
+$Self->DoneTesting();
```
bschmalhofer: can be replaced with `done_testing;`

scripts/test/Ticket/TicketMasks.t
```diff
@@ -0,0 +1,300 @@
+# --
+# OTOBO is a web-based ticketing system for service organisations.
+# --
+# Copyright (C) 2001-2020 OTRS AG, https://otrs.com/
```
bschmalhofer: I think that for new files there is no copyright for OTRS.

bschmalhofer (approved):
Thanks for considering the suggestions.

PR #4675, opened 2025-09-09
4599 left margin in input field content in dynamic field sets with grids

### LINKED ISSUE #4597: Label widths in set dynamic fields.
Currently labels in sets have a fixed width of 25%. This makes sense outside of grids, but does not work well with grids, where the width should be 100% (of the grid cell). This was found for the CMDB but is believed to be the same for the core. Please have a look at both.

### LINKED ISSUE #4599: Left margin in input field content in dynamic field sets with grids.
At least text input fields do not have any margin/padding on the left, if they are in grids in a set (possibly also in grids outside of sets). Please compare with normal text fields on input masks:
Image

chzauleck:


MichaelThumes (approved):
looks good to me

Issue #4554, opened 2025-07-31
Consider making exec_web() in entrypoint.sh ENV driven


MichaelThumes:
The exec_web() function in entrypoint.sh (see [here](https://github.com/RotherOSS/otobo/blob/rel-11_0/bin/docker/entrypoint.sh#L103)) currently has some exciting comments on how to modify it for development purposes. However to try out these options modification of entrypoint.sh and a local image build is necessary.

To avoid unnecessary custom image builds, please consider making exec_web() more ENV variable driven, and switch the way of starting otobo web server via 'exec plackup' based on an ENV variable.

Rough Idea (here for rel-10_0):

```Diff
diff --git a/bin/docker/entrypoint.sh b/bin/docker/entrypoint.sh
index 53f10c3a5..08efa1c63 100755
--- a/bin/docker/entrypoint.sh
+++ b/bin/docker/entrypoint.sh
@@ -101,17 +101,21 @@ function stop_daemon() {
 # Start the webserver
 function exec_web() {
 
-    # For development omit the --env option, thus setting PLACK_ENV to its default value 'development'.
-    # This enables additional middlewares that are useful during development.
-    # For development also enable the -R option. This watches for changes in the modules and the config files.
-    # otobo.psgi is watched implicitly.
-    #   exec plackup --server Gazelle -R Kernel --port 5000 bin/psgi-bin/otobo.psgi
-
-    # For debugging reload the complete application for each request by passing -L Shotgun
-    #   exec plackup -L Shotgun --port 5000 bin/psgi-bin/otobo.psgi
-
-    # For production use Gazelle, which is implemented in C
-    exec plackup --server Gazelle --env deployment --port 5000 bin/psgi-bin/otobo.psgi
+    if [ "$OTOBO_DEVEL" = "devel" ]
+    then
+        # For development omit the --env option, thus setting PLACK_ENV to its default value 'development'.
+        # This enables additional middlewares that are useful during development.
+        # For development also enable the -R option. This watches for changes in the modules and the config files.
+        # otobo.psgi is watched implicitly.
+        exec plackup --server Gazelle -R Kernel --port 5000 bin/psgi-bin/otobo.psgi
+    elif [ "$OTOBO_DEVEL" = "debug" ]
+    then
+        # For debugging reload the complete application for each request by passing -L Shotgun
+        exec plackup -L Shotgun --port 5000 bin/psgi-bin/otobo.psgi
+    else
+        # For production use Gazelle, which is implemented in C
[...]

PR #4683, opened 2025-09-11
Issue #4554: first stab at flexible webserver

bschmalhofer:
the command 'web' now supports the modifiers
- deployment (the default)
- development
- shotgun


MichaelThumes (approved):
changes look good to me for rel-10_0.

pls take care when upmerging, as exec_web() has changed (S3 support, if I recall correctly)

Issue #2297, opened 2023-05-05
Errors and warnings emitted by the test suite during global destruction


bschmalhofer:
When running the test suite I sometimes see messages like:

`[Thu Apr 13 13:14:01 2023] TicketCreateIncludeTicketData.t: Use of uninitialized value $alias in lc at /usr/local/lib/perl5/5.32.1/x86_64-linux-gnu/Encode/       Alias.pm line 38 during global destruction.
`
These messages occur in rel-10_0, rel-10_1 and rel-11_0. They do not seem to be related to test errors. I can't tell whether they indicate a real problem.

bschmalhofer:
There are still a bunch of warning emitted during global destruction.
```
$ grep -r 'during global destruction' /home/bernhard/devel/OTOBO/docker/proofs/prove_11.0.x_2024-06-06-122436_rel-11_0.out | wc -l
76
```
Not sure how many of these are harmless.

bschmalhofer:
This has actually improved:

```
grep -r 'during global destruction' /home/bernhard/devel/OTOBO/docker/proofs/prove_11.0.x_2024-11-15-100117_rel-11_0.out | wc -l
11
```

bschmalhofer:
Noticed this again when working on #4452 . It would be nice if these kind of warnings could be reduced.

bschmalhofer:
Encountered a new warning during global destruction:

> Use of **uninitialized value $^V in sprintf** at /opt/otobo/Kernel/System/Log.pm line 297 during global destruction.
> ERROR: ?LogPrefix?-83 **Perl:**  OS: linux Time: Thu Sep 11 13:34:56 2025
> 
>  Message: 8000-GenericInterface died with: Unknown encoding 'UTF-8' at /opt/otobo/Kernel/System/Cache/Redis.pm line 120 during global destruction.
> 
> 
>  Traceback (3869): 
>    Module: Kernel::System::EventHandler::EventHandler Line: 245
>    Module: Kernel::System::EventHandler::EventHandlerTransaction Line: 320
>    Module: Kernel::System::Ticket::DESTROY Line: 8423
>    Module: (eval) Line: 615
>    Module: Kernel::System::ObjectManager::ObjectsDiscard Line: 615
>    Module: Kernel::System::ObjectManager::DESTROY Line: 677

[...]

PR #4714, opened 2025-09-18
Issue #2297: discard objects in $Kernel::OM before global destruction

bschmalhofer:
The is done by registering an END handler in K::S::UnitTest::RegisterOM. This lets the test scripts behave more like the other OTOBO code.

Thus the special check, regarding 5^V, is no longer necessary in K::S::Log.


bschmalhofer:
Ignoring 

> Miscellanea::ProhibitUselessNoCritic violated at Kernel/System/Log.pm 369 column 30 (Severity: 4)

Issue #4405, opened 2025-06-13
Code Blocks are not displayed in Monospace


73:
### Environment

- OTOBO version: 11.0.10

### Expected behavior

I expect code blocks to be presented in a monospace font.

### Actual behavior

The font is not changed in a code block and can not be changed within code blocks.

### How to reproduce

Well set a code block.

### Additional information

This also happens in the FAQ and the customer/public portal.
It would be nice to see code in a monospaced font. Ideally, the same font for `inline code` and ...
```
code blocks
```
And It should go well with the default font and Quicksand.
The Courier font looks out of place (see screenshot).
 
### Screenshots

<img width="1263" alt="Image" src="https://github.com/user-attachments/assets/4d847d0f-1075-479a-a166-6d200d278119" />

svenoe:
In the styles there is a rule
```
code {
  font-family: monospace;
}
```
which is overridden however by
```
.ck-content * {
  font-family: [...]
}
```

Adding `font-family: monospace;` in a different existing rule
```
.ck-content code {}
```
or probably rather just adding the `.ck-content` to the first rule, because we do not want ck editor to affect elements outside of ck, should solve this issue.

PR #4715, opened 2025-09-18
Issue #4405: adapted EmailParser test to reflect changes in css

chzauleck:


bschmalhofer (approved):
Looks good to me. Only the new CSS rule had to be adapted.

PR #4732, opened 2025-09-29
clarify security notes

73:
I did not find a "Security Researcher Hall of Fame", however, we credit those reports in the release notes.


bschmalhofer (commented):
That should be "release notes"

73:
@bschmalhofer thank you. Good catch. I also applied the boy scout rule and fixed little the wording and format.

bschmalhofer:
The list of the  supported release series should be updated too in _SECURITY.md_ .

PR #4734, opened 2025-09-30
wp #1213: fix IsHTMLContent initial value for RichText

rmfalves:


MichaelThumes:
testsuite was run against the dev branch and there were no errors reported:
https://git.otobo.org/otobo/otobo/-/pipelines/47417/test_report?job_name=test_suite

PR #4742, opened 2025-10-02
Issue #4684: Check if dynamic field isn't assigned as deactivated.

### LINKED ISSUE #4684: DynamicFieldWidget is visible when DynamicFields are assigned (even when all are de-activated)
When adding DynamicFields via DynamicField Screens as de-activated to DynamicFieldWidgetDynamicField, then in AgentTicketZoom the Widget is visible, even when there is no DF assigned as activated.

In my opinion it makes sense to only show the widget when a DF is assigned as activated and preferably only when a value is shown and therefore the field is there.

OTOBO 11.0.11 Docker

stefanhaerter:
Until now, if a dynamic field was assigned as deactivated element to the
dynamic field widget but had a value for the ticket, the widget would be
displayed, but empty.

An additional check line was added to prevent this.

closes #4684


MichaelThumes (approved):
looks good to me

Issue #1849, opened 2022-07-20
Kernel::System::DynamicField::Backend::RandomValueSet() does not support ObjectName


bschmalhofer:
My expectation would be that Kernel::System::DynamicField::Backend::RandomValueSet() behaves mostly like  Kernel::System::DynamicField::Backend::ValueSet(), only that a random value is used. But RandomValueSet() does not support the parameter `ObjectName`.

Furthermore the design is a bit strange. It is fine that that dynamic field classes have their specific random values. But then I propose that the driver have a function that only generated the random value. In this way Kernel::System::DynamicField::Backend::RandomValueSet()  can simply call Kernel::System::DynamicField::Backend::ValueSet() internally. 

This is not urgent, as it looks like `RandomValueSet()` is only used in the console command `Dev::Tools::Database::RandomDataInsert`

stefanhaerter:
As far as I understand it, https://github.com/RotherOSS/otobo/blob/f720e81f0043f38fb570b6705d7932cbf9ac36c5/Kernel/System/DynamicField/Backend.pm#L2060 calls the RandomValueSet()-function of the respective backend, and at least BaseText calls its ValueSet(): https://github.com/RotherOSS/otobo/blob/f720e81f0043f38fb570b6705d7932cbf9ac36c5/Kernel/System/DynamicField/Driver/BaseText.pm#L607

I haven't checked the other backends yet.

rmfalves:
Regarding @stefanhaerter's point, I checked a bit further on the CheckBox, Date and DateTime backends, and they also call the ValueSet() in the last lines of RandomValueSet():
```
    my $Success = $Self->ValueSet(
        %Param,
        Value => $Value,
    );

    if ( !$Success ) {
        return {
            Success => 0,
        };
    }
    return {
        Success => 1,
        Value   => $Value,
    };
```

rmfalves:
The `RandomValueSet` and the `ValueSet` functions clearly share many common patterns in their code, with the former being almost a subset of the latter, suggesting that @bschmalhofer has a valid point.

However, I have only modified the function `Kernel::System::DynamicField::Backend::RandomValueSet()` to handle the `ObjectName` the same way as `ValueSet()`. Implementing what @bschmalhofer suggested requires a careful refactoring, which in my opinion is beyond the scope of this issue.

So, I have opened a new issue on that matter.

svenoe:
[...]

PR #4749, opened 2025-10-07
Issue #1849: Added ObjectName parameter to RandomValueSet

rmfalves:


svenoe:
Might have a use at some other point in time; see also the issue.

Issue #4572, opened 2025-08-07
Set dynamic fields with Grids lead to Internal Server Error in AgentTicketSearch and Ticket Overviews


stefanhaerter:
Using Set dynamic fields with Grids in its Include-Structure leads to an Internal Server Error in AgentTicketSearch and Ticket Overviews. Stacktrace is:

```
ERROR: OTOBO-CGI-60 Perl: 5.38.4 OS: linux Time: Thu Aug  7 07:57:07 2025

 Message: Need ID or Name!

 RemoteAddress: 46.223.48.142
 RequestURI: /otobo/index.pl

 Traceback (26493):
   Module: Kernel::System::DynamicField::DynamicFieldGet Line: 256
   Module: Kernel::Modules::AgentTicketSearch::Run Line: 229
   Module: Kernel::System::Web::InterfaceAgent::Content Line: 1282
   Module: Kernel::System::Web::InterfaceAgent::Response Line: 1364
   Module: Kernel::System::Web::App::call Line: 86
   Module: Plack::Component::__ANON__ Line: 50
   Module: Plack::App::URLMap::call Line: 71
   Module: Plack::Component::__ANON__ Line: 50
   Module: Plack::Sandbox::_2fopt_2fotobo_2fbin_2fpsgi_2dbin_2fotobo_2epsgi::__ANON__ Line: 188
   Module: Plack::Sandbox::_2fopt_2fotobo_2fbin_2fpsgi_2dbin_2fotobo_2epsgi::__ANON__ Line: 158
   Module: Plack::Sandbox::_2fopt_2fotobo_2fbin_2fpsgi_2dbin_2fotobo_2epsgi::__ANON__ Line: 144
   Module: Plack::Middleware::HTTPExceptions::try {...}  Line: 20

Can't use an undefined value as a HASH reference at /opt/otobo/bin/psgi-bin/../../Kernel/Modules/AgentTicketSearch.pm line 232.
```

The relevant place in the code is:

https://github.com/RotherOSS/otobo/blob/d6da5cb110c2ffc9937070f4194e3e24091208d8/Kernel/Modules/AgentTicketSearch.pm#L218-L237

svenoe:
(Also keep the CMDB in mind, please.)

Issue #4752, opened 2025-10-08
React to nginx:mainline Docker image being upgraded to Debian 13 Trixie


bschmalhofer:
The Docker Image nginx:mainline, https://hub.docker.com/layers/library/nginx/mainline/images/sha256-5f217ea1c23175f517f39c94f498e35bfa92e82627aaa61b359c8ab1db065f14, has been updated to use Trixie. See https://github.com/nginx/docker-nginx/pull/995 . This has repercussions in building the OTOBO specific nginx Docker image. This surfaced in the DockerImageUpdateAutobuild workflow, https://github.com/RotherOSS/otobo/actions/runs/18332333304 .

The error when building was:
```
#8 1.742 Reading state information...
#8 1.753 Package libpcre3-dev is not available, but is referred to by another package.
#8 1.753 This may mean that the package is missing, has been obsoleted, or
#8 1.753 is only available from another source
#8 1.753 
#8 1.755 E: Package 'libpcre3-dev' has no installation candidate
#8 ERROR: process "/bin/sh -c apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-recommends install        gcc         libc-dev         make         libpcre3-dev         zlib1g-dev         libkrb5-dev         wget" did not complete successfully: exit code: 100

```
TODO:

- [x] get the build running again
- [x] upmerge from rel-10_0 to rel-11_1

bschmalhofer:
See also #4507

bschmalhofer:
libpcre3 is indeed gone from Debian 13 Trixie,  https://bugs-devel.debian.org/cgi-bin/bugreport.cgi?bug=1071970 . Confusingly it is replaced with libpcre2, https://linux.debian.devel.narkive.com/s8mlB5Xw/pcre-package-naming .

bschmalhofer:
Closing this issue.

But keep in mind to revert the changes in _.github/workflows/docker_image_update_autobuild.yml_ after the next patch level release. 
Also __.github/workflows/docker_image_update_all.yml_ might need to skip the Nginx Images when it is invoked next.

bschmalhofer:
Discussed this at an onsite team meeting and decided to safeguard against future Debian version updates. So pin the Debian version and use _mainline-trixie_.
While at it, install build-essentials instead of the specific package list.

bschmalhofer:
Done.
[...]

PR #4753, opened 2025-10-08
Issue #4572: use libpcre2 instead of libpcre3

bschmalhofer:
This is because nginx:mainline switched to Debian 13 Trixie. And Trixie no longer ships libpcre3, only the newer libpcre2. Yes, libpcre2 is newer than libpcre3


bschmalhofer:
Sorry, misspelled again, this PR belongs to #4752

PR #4782, opened 2025-10-21
Issue #4774: remove invalid input from autocomplete reference fields

### LINKED ISSUE #4774: invalid input in Autocomplete reference fields should be removed
when typing into a reference dynamic field with autocomplete, invalid values still in the field should be automatically deleted on leaving the field. this prevents confusion on the fact that invalid input will not be saved when the form is submitted.

chzauleck:


MichaelThumes (approved):
looks good to me

PR #4796, opened 2025-10-28
Issue #4432: Removed ITSM-related unused methods.

### LINKED ISSUE #4432: Clean up ITSM ConfigItem related methods in Kernel::System::ZnunyHelper
In _ZnunyHelper.pm_ there are methods like `_ITSMConfigItemVersionAdd()` . These seem to be obsolete for at least two reasons:

- OTOBO core should not be concerned about config items
- these method still use the outdated XML Data structure

- [ ] Collect the relevant methods
- [ ] Check whether these methods are used in OTOBO core
- [ ] Check whether these methods are use in ITSM ConfigurationManagement and related management
- [ ] Check whether migration scripts use these methods
- [ ] Check whether there is code that is worth to keep 
- [ ] get rid of it

I propose to do this already in rel-11_0.

stefanhaerter:
Also cleaned up some unused variables.

closes #4432


MichaelThumes (approved):
looks good to me

PR #4801, opened 2025-10-30
Issue #4800: Ensures tables span full width in rich text editor

### LINKED ISSUE #4800: Change default styling of tables in RichText
The default style of tables in RichText should be:
centered
width 100%

chzauleck:
Sets default table width to 100% in both editor configuration and article content styling.


MichaelThumes (approved):
this looks 100% good to me

PR #4811, opened 2025-11-06
Issue #4809: Added missing cache key filling.

### LINKED ISSUE #4809: Ticket event TicketDynamicFieldDefault does not actually prevent multiple execution
The event TicketDynamicFieldDefault attempts to prevent multiple execution per ticket by setting and evaluating a cache key on the ticket object, but the key is not set in successful execution case, thus allowing multiple executions.

stefanhaerter:
closes #4809


MichaelThumes (approved):
Looks good!

PR #4813, opened 2025-11-06
Removed overlooked comment.

stefanhaerter:


bschmalhofer:
I grepped for other occurences of 'EO ' and found a faulty comment in IO::SessionData.

```
grep -r 'EO ' Kernel/
Kernel/cpan-lib/PDF/API2/Resource/XObject/Form/BarCode/ean13.pm:my @parity = qw(OOOOOO OOEOEE OOEEOE OOEEEO OEOOEE OEEOOE OEEEOO OEOEOE OEOEEO OEEOEO);
Kernel/cpan-lib/IO/SessionData.pm:# 0EO ("0 but true") if the read would block, and undef on EOF and other failures.
```

IO::SessionData is only in _Kernel/cpan-lib_ because SOAP::Lite needs it. Both distros look very much unmaintained. Can't even file an issue in the RT-Bugtracker. Sometimes the Perl ecosystem is frustrating indeed.

Issue #4836, opened 2025-11-13
No need to check for the requirement  Net::SMTP


bschmalhofer:
Net::SMTP is a Perl core module since a long time. See http://perlpunks.de/corelist/version?module=Net%3A%3ASMTP . Thuse there is no need to check for it. A quick search did not indicate that any Linux distribution removes Net::SMTP from their Perl installation. 

See also #4834

bschmalhofer:
Done, closing this issue.

PR #4837, opened 2025-11-13
Issue #4836: do not check for core module Net::SMTP

bschmalhofer:
thus the feature 'mail' can also be removed


bschmalhofer:
Ignoring the complaint about FindBin again.

Issue #4833, opened 2025-11-13
Allow to activate profiling in the web app via the .env file


bschmalhofer:
> Currently profiling the web app with Devel::NYTProf is done by manually editing _bin/psgi-bin/otobo.psgi_ . There the variable `my $ProfilingIsActive = 1;` needs to be set. This is fine for local installations. Especially for autmated testing within GitLab or GitHub it would be nice if profiling could be activated from the outside. The mechanism from #4554 , an additional parameter to _entrypoint.sh_ can be used for that.
> 
> This should be targeted for rel-10_1, as that is the first release series that uses the standard middleware for profiling.
>  

 _Originally posted by @bschmalhofer in [#3769](https://github.com/RotherOSS/otobo/issues/3769#issuecomment-3527530107)_

bschmalhofer:
It would also be nice if the directory _var/httpd/htdocs/nytprof_ would be served by https://metacpan.org/pod/Plack::App::DirectoryIndex . This is convenient, because when cĺicking on a directory that contains _index.html_ the HTML page is shown, not the directory listing with potentially many files.

bschmalhofer:
See also https://ideas.otobo.io/posts/27/provide-a-mod_autoindex-emulation-in-otobo-psgi. Go Vote!

bschmalhofer:
We discussed this internally. The decisions were:

- _entrypoint.sh_ will be enhanced so that profiling can be activated in _.env_
- no new required dependencies will be added to the native installation or the Docker image
- the profiling results will be placed in otobo-web/static/nytprof
- otobo-web/static will be served as static files, can be used e.g. for quickly publishing static files, e.g. documentation and reports

This means that for actual profiling the modules 

This means that for profiling  'Devel::NYTProf and Plack::Middleware::Profiler::NYTProf must be installed manually into _/opt/otobo/local_.

Plack::App::DirectoryIndex is not available as a Debian package. So let's stick with Plack::App::Directory instead.

bschmalhofer:
This has been tested manually and it looks mostly sane.

One finding is that `OTOBO_WEB_OPTION=nytprof` should only be activated in _.env_ after the needed modules have been installed in _/opt/otobo/local_ . 

Closing this issue.

PR #4838, opened 2025-11-13
Issue #4833 nytprof

bschmalhofer:


bschmalhofer:
@svenoe : Can you take a look at this? I tend to generally install Devel::NYTProf in the image. The number of dependencies is lower than I expected. See   https://github.com/RotherOSS/otobo/compare/rel-10_1...issue-%234833-nytprof?expand=1#diff-fc00e6ffb3c763013c9e771e6f98e462f03770f325ecfb81ec8408a8d663d47b .

cpanfile.docker
```diff
@@ -141,6 +141,15 @@ requires 'Const::Fast';
 
 # };
 
+# feature 'devel:profiling', 'Modules for profiline Perl code' => sub {
```
bschmalhofer: "profiline" should be "profiling"

PR #4841, opened 2025-11-17
Issue #4823: Improve completion by appending  / to directories.

### LINKED ISSUE #4823: Improve bash autocompletion to put a '/' after completing a directory
As a fallback, the bash completion console command uses the list of *files* in the current directory:

https://github.com/RotherOSS/otobo/blob/429e6b7c5bc90f2bbe14b995680150d540d07047/Kernel/System/Console/Command/Internal/BashCompletion.pm#L106-L111

This is unpleasant if one wishes to autocomplete a directory. It would be nice if we could replace `glob` with something to list files and directories alike.

stefanhaerter:
closes #4823


stefanhaerter:
Closing due to low relevance.

PR #4862, opened 2025-11-20
Fixed typo.

stefanhaerter:


bschmalhofer:
I think that 'compliant' is the better word in this context.

PR #4876, opened 2025-11-24
 Issue #4391: Replaced TicketSearch call with number of TicketIDs. 

### LINKED ISSUE #4391: Eliminate unneeded TicketSearch in AgentTicketStatusView.pm
In https://github.com/RotherOSS/otobo/blob/e2ee2127322b368d435365fbd0f878e7d8d7c055/Kernel/Modules/AgentTicketStatusView.pm#L248 there is a TicketSearch() that gets the number of viewable tickets. There is no obvious reason why this search is necessary. The other ticket list don't have that search as the list of viewable tickets is already available.

stefanhaerter:
closes #4391

### DIFF
--- a/Kernel/Modules/AgentTicketStatusView.pm
+++ b/Kernel/Modules/AgentTicketStatusView.pm
@@ -245,11 +245,7 @@ sub Run {
             Result => 'ARRAY',
         );
 
-        $ViewableTicketCount = $TicketObject->TicketSearch(
-            %{ $Filters{$Filter}->{Search} },
-            %ColumnFilter,
-            Result => 'COUNT',
-        ) || 0;
+        $ViewableTicketCount = scalar @ViewableTickets;
     }
 
     my $View = $ParamObject->GetParam( Param => 'View' ) || '';


Kernel/Modules/AgentTicketStatusView.pm
```diff
@@ -245,11 +245,7 @@ sub Run {
             Result => 'ARRAY',
         );
 
-        $ViewableTicketCount = $TicketObject->TicketSearch(
-            %{ $Filters{$Filter}->{Search} },
-            %ColumnFilter,
-            Result => 'COUNT',
-        ) || 0;
+        $ViewableTicketCount = scalar @ViewableTickets;
```
bschmalhofer, line 248: @stefanhaerter , did you verify that the behavior is the same when `$ViewableTicketCount` now comes from a search that has a `Limit => $Limit` ?

stefanhaerter: Not yet - I'll try to do so as quickly as possible

stefanhaerter:
A brief discussion showed that further research is necessary. Closing the PR.

Issue #4881, opened 2025-11-26
Include the Perl module Encode::HanExtra in the Docker image


bschmalhofer:
Currently the Perl module [Encode::HanExtra](https://metacpan.org/pod/Encode::HanExtra) is installed only for native installations. This module adds support for less used chinese encodings. It is not obvious whether these encodings are still relevant. But there is no good reason why there should be a disctiction between Docker and native installations.

TODO:

- [x] install Encode::HanExtra also in the Docker based installations
- [ ] Add a test case that actually uses that encoding

The milestone is set to rel-10_1 as this is the first release series that hase `%IsCommonFeature` in _bin/otobo.CheckModules.pl`.

bschmalhofer:
Actually testing different encodings in incoming mails looks like a good thing. So handle this in a seperate issue.

bschmalhofer:
Encode::HanExtra has been added. Adding a test is covered in a new issue. Closing this issue.

PR #4882, opened 2025-11-26
Issue #4881 han extra

bschmalhofer:


bschmalhofer:
Ignoring the complaint about FindBin.

PR #4907, opened 2025-12-09
Issue #3422: Optimized regex and standardized inline detection

### LINKED ISSUE #3422: Rewrite detect attachment postmaster filter.
The current implementation hosts a very inefficient regex and can cause emails with multiple large HTML attachments to consistently run into timeouts. It is disabled in the standard configuration for OTOBO 11.0 currently (see https://github.com/RotherOSS/otobo/issues/3419), but should ideally be rewritten.

stefane2741:
closes #3422


MichaelThumes (approved):
looks good to me

PR #4925, opened 2025-12-10
Issue #4755: Prevent nowrap for title dynamic fields.

### LINKED ISSUE #4755: Long DynamicField Title text is being cut off in AgentTicketPhone
When having a DF type "Title" you can put in texts as a hint for different forms.
When having such a field in e.g. AgentTicketPhone and maybe having a bit of a zoom situation, the text is just being cut off:
Image

I expect the behavior to be similar to the customer frontend (automatic line breaks). In CustomerTicketMessage it looks like this:
Image

Tested on OTOBO 11.0.12 with macOS and Safari

stefanhaerter:
closes #4755


stefanhaerter:
Closed in favor of #4926

PR #4932, opened 2025-12-11
Issue #4931: Pass customer and customeruser ids on.

### LINKED ISSUE #4931: Reference dynamic fields: Matching CustomerID and / or CustomerUserID doesn't work in CustomerTicketMessage
Matching reference dynamic field items via CustomerID or CustomerUserID does only work on initial page rendering in CustomerTicketMessage. This is due to the fact that CustomerID and CustomerUserID are not passed on during the AJAX update call to GetFieldStates().

stefanhaerter:
The CustomerID and CustomerUserID need to be passed on to enable reference dynamic fields to match on them on AJAX update.

closes #4931


stefanhaerter:
Closing in favor of #5308

Issue #3813, opened 2024-09-24
Increase unit test coverage for FieldRestrictions.pm


MichaelThumes:
This scenario lacks test coverage:

```Perl
# get values and visibility of dynamic fields
                %CurFieldStates = $FieldRestrictionsObject->GetFieldStates(...)
```

stefanhaerter:
Can this be considered done, taking into account the work on the test within #4547?

PR #4955, opened 2025-12-17
Issue #3813: cleanup ACLS before test run

MichaelThumes:


bschmalhofer (commented):
Maybe add a couple of checks, so that we don't end up with a confusing empty subtest

Issue #4979, opened 2025-12-20
Remove HTTP::Message from Kernel/cpan-lib for OTOBO 10.1.x


bschmalhofer:
> There were no problems encounterd in rel-11_0. So this cleanup can safely also be done for rel-10_1. As before, the goal is to eliminate divergences between the release series. 

 _Originally posted by @bschmalhofer in [#2612](https://github.com/RotherOSS/otobo/issues/2612#issuecomment-3677706548)_

bschmalhofer:
Cherry-picking the commits  #2821328 and #8de293d .

bschmalhofer:
the test suite showed no problems, PR is merged, closing this issue

bschmalhofer:
Reverting this change as there should be no interface changes in patch level releases.

bschmalhofer:
Done. Closing this issue again.

PR #4980, opened 2025-12-21
Issue #4979 http message

bschmalhofer:


bschmalhofer:
Ignoring the complaint about FindBin.

PR #4993, opened 2026-01-05
Fixed typos.

urbalazs:
Fixed some typos.

Can you cherry-pick these fixes to other supported branches or should I create separate pull requests for branch `rel-10_1` and `rel-11_1`?


bschmalhofer:
I took a quick look and the changes look sensible to me. However there are quite a few changes and my gut feeling is that this PR should be split up. For example there are changes in code comments and at least one interface change in `Kernel::System::UnitTest`. Interface changes should be done only in new minor releases.

Could you provide a PR for rel-10_1 with changes only in code comments? This makes it easier for is in reviewing the more significant changes.

svenoe:
Partly integrated in https://github.com/RotherOSS/otobo/issues/5064

PR #4999, opened 2026-01-05
Fixed layout in agent personal preferences, standardized wording.

urbalazs:
This PR standardizes wording by using sentence case in agent personal preferences.

The `Key` is removed for each setting to have the same look and feel. Here is a before-after screenshot:

agent-personal-preferences-before-after


svenoe:
Partly integrated in https://github.com/RotherOSS/otobo/issues/5064

PR #5019, opened 2026-01-09
Improved close button label.

urbalazs:
In some languages (e.g. in Hungarian), there are different word for closing a ticket and closing a dialog. This PR makes precise wording so that translators can add the right translation.

This is a consistency change, since any other *Close* button uses the same wording.


svenoe:
Hi Balázs,
quickly here - thanks for all the prs; since many of them, like this one, are not technical changes, I will go through them together with a colleague this week and decide on what to merge.
Best regards, Sven

svenoe:
Thanks - one comment, please use otobo.io instead of otobo.de; probably it was a code policy correction, I will have a look at the versions there.

PR #5027, opened 2026-01-13
Issue #4712: Made MultipartMixed.t ckeditor-CSS independent.

### LINKED ISSUE #4712: non breaking changes to RichText css when parsing email
in some circumstances when parsing an email, plain text email content is (among other things) converted to html using   DocumentComplete()

the resulting css added to the email, does however vary slightly from what is actually found in the relevant files.

this has been observed in `EmailParser/MultipartMixed.t:l252`, where there is an added `\n.ck-content {}` in the parsed email, that is not part of the original css file (`var/httpd/htdocs/skins/Agent/default/css/RichTextArticleContent.css`)

stefanhaerter:
Changes to the ckeditor CSS should no longer break the test.

Only exception is the file size.

closes #4712


stefanhaerter:
Rejected, closing the PR.

Issue #1860, opened 2022-07-29
Preparations for switch from DBD::mysql to DBD::MariaDB


bschmalhofer:
My hunch is that DBD::MariaDB is a better interface for MySQL and MariaDB than DBD::mysql. It should be investigated whether it makes sense to switch. See also #437.

bschmalhofer:
Let's test this, whether this is much effort.

bschmalhofer:
If that ever happens then the special handling for the '5.5.5-' prefix can be removed in _Kernel/Modules/Installer.pm_.

bschmalhofer:
I have looked into this issue again. According to https://metacpan.org/release/DVEEDEN/DBD-mysql-4.050/view/lib/DBD/mysql/INSTALL.pod and https://metacpan.org/dist/DBD-MariaDB/view/lib/DBD/MariaDB/INSTALL.pod, both drivers support connection to MySQL and to MariaDB servers using either MySQL or MariaDB client libraries. Any combination is possible. 

In the standard Docker installation the database is MariaDB. The webserver runs in a Debian based image with the package _default-mysql-client_ installed. This is actually the MariaDB native client.

_bin/otobo.CheckModules.pl_ suggests, for Debian based systems,  to install _libdbd-mysql-perl_ which depends on _libmariadb3_ . I did not check the other Linux distributions. 

The plan is to test with DBD::MariaDB 1.23 first. When there are no showstoppers we need to get back and check the availability of DBD::MariaDB in the supported systems.

Update: DBD::mysql >= 5.001 does not support building against MySQL 5.7  and MariaDB client libraries. DBD::MariaDB is now at version 1.24

bschmalhofer:
I tried migrating to DBD::MariaDB and found that double encoding happens for LONGBLOB columns. I enhanced the test script  _scripts/test/DB/UTF8Range.t_ in order to have a simple test case for the double encoding. Let's find out whether there is a simple fix for that. If the fix is not simple then we simply refrain from using DBD::MariaDB.

See also: https://perladvent.org/2018/2018-12-08.html

bschmalhofer:
[...]

PR #5069, opened 2026-01-22
Issue #1860 dbd mariadb based on 11 1

bschmalhofer:


Kernel/System/DB.pm
```diff
@@ -153,6 +158,7 @@ sub new {
 
         # otherwise auto detection from the DSN
         return 'mysql'      if $Self->{DSN} =~ m/:mysql/i;
```
svenoe: `m/:mysql/i` should not be possible anymore, after the substitution on line 130. (Do we want to return "mariadb" instead of mysql? Maybe/Probably not, did not look further into it.)

Kernel/System/DB.pm
```diff
@@ -597,8 +622,29 @@ sub Do {
 
     return unless $Self->Connect;
 
-    # send sql to database
-    if ( !$Self->{dbh}->do( $Param{SQL}, undef, @Array ) ) {
+    # send sql to database, taken from the implemetation of DBI::do
```
svenoe, line 624: I'm a bit surprised that there is no way of letting DBI do this within the `do` somehow...

Kernel/System/DB.pm
```diff
@@ -597,8 +622,29 @@ sub Do {
...
+    my $StatementHandle = $Self->{dbh}->prepare( $Param{SQL} );
+
+    return unless $StatementHandle;
```
svenoe: Is there no `$DBI::errstr` (see line 648ff below), which we should log here and also in line 644?

Kernel/System/DB.pm
```diff
@@ -597,8 +622,29 @@ sub Do {
...
+        continue {
+            $Index++;    # bind index are based on 1
+        }
+    }
+    my $ExecuteSuccess = $StatementHandle->execute(@Array);
+
+    return unless $ExecuteSuccess;
+
+    my $Rows = $StatementHandle->rows;
+    if ( $Rows != 0 && !$Rows ) {
```
svenoe: Is this from `DBI::do`? Looks strange, why not `if ( !defined $Rows ) {` or maybe `if ( !defined $Rows || $Rows eq '' ) {` if the latter is possible?

Kernel/System/DB/mysql.pm
```diff
@@ -91,9 +93,10 @@ sub LoadPreferences {
     #$Self->{'DB::ShellConnect'} = '';
 
     # init sql setting on db connect
-    if ( !$Kernel::OM->Get('Kernel::Config')->Get('Database::ShellOutput') ) {
-        $Self->{'DB::Connect'} = 'SET NAMES utf8mb4';
-    }
+    # TODO: test whether this is needed for MySQL
+    #if ( !$Kernel::OM->Get('Kernel::Config')->Get('Database::ShellOutput') ) {
+    #    $Self->{'DB::Connect'} = 'SET NAMES utf8mb4';
+    #}
```
svenoe, line 99: Should not be forgotten I guess.

Kernel/System/DynamicFieldDB.pm
```diff
@@ -92,10 +94,17 @@ sub new {
...
     }
     else {
-        $DatabaseDSN = 'DBI:' . $Self->{DynamicFieldConfig}->{Config}->{DBType}
-            . ':database=' . $Self->{DynamicFieldConfig}->{Config}->{DBName}
-            . ';host=' . $Self->{DynamicFieldConfig}->{Config}->{Server}
-            . ';port=' . $Self->{DynamicFieldConfig}->{Config}->{Port};
+
+        # The driver module DBD::MariaDB expects an integer value for the port when given.
+        # Therefore set the port only when there actually is a port.
+        $DatabaseDSN = join ':',
```
svenoe: Careful - host and port are concatenated using a semicolon, not a colon in the old code. Is this an intentional change?

Kernel/System/MailQueue.pm
```diff
@@ -1377,18 +1377,19 @@ sub _CheckValidMessageData {
...
 sub _SerializeMessage {
     my ( $Self, %Param ) = @_;
 
-    my $DBObject       = $Kernel::OM->Get('Kernel::System::DB');
     my $StorableObject = $Kernel::OM->Get('Kernel::System::Storable');
     my $Message        = $StorableObject->Serialize(
         Data => $Param{Message},
     );
+
+    my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
```
svenoe: Minor, but initializing the necessary "Objects" via Kernel::OM in one bundled place, if there is no possible return or a lot of unconnected code in between is very OTOBO like, so I don't see a reason for moving this and enlarging the diff.

svenoe (commented):
I did not check the changes to the migration, I believe we should remove it. Will discuss that internally. Also I'm not sure about the encoding, but I think you know what you are doing here. From what I saw, in every place the data was directly given in `%Param` and not as reference, where we would have to check whether it is used further by the calling method. I also did not check the otobo schema where other binary columns are; do we have to do that, or can we be certain that the DirectBlob in the code gives us every relevant instance?

bschmalhofer:
> I also did not check the otobo schema where other binary columns are; do we have to do that, or can we be certain that the DirectBlob in the code gives us every relevant instance?

The only binary colums are apparently LONGBLOB columns. I started to check these columns individually until I saw the pattern with 'DirectBlob'. The idea is that all relevant locations are covered there, because otherwise the PostgreSQL backend would have problems there. Assuming here that the PostgreSQL backend is fine as there are productive installations using PostgreSQL.

Kernel/System/DB.pm
bschmalhofer: > `m/:mysql/i` should not be possible anymore, after the substitution on line 130. (Do we want to return "mariadb" instead of mysql? Maybe/Probably not, did not look further into it.)

Good catch. 

I was thinking about renaming to 'mariadb' too. But that would entail to rename _Kernel/System/DB/mysql.pm_ to mariadb.pm. I wanted to avoid that.

Kernel/System/DB.pm
bschmalhofer: Yes, I was surprised too.

Kernel/System/DB.pm
bschmalhofer: Yes, that is verbatim from DBI::do. I decided to keep it like I saw it there, as there might be a reason for that.

Kernel/System/DB.pm
bschmalhofer: Very good catch. The error message would be lost in that case. I'll wrap those lines of code  in an block labelled with `DBI_DO_EMULATION`. Leaving that block with `next` is then comparable  to `return` in a sub.

Kernel/System/DB/mysql.pm
bschmalhofer: Yes, the DBD::MariaDB says that it is not needed for both MariaDB and MySQL. But testing with MySQL must still be done.

Kernel/System/DynamicFieldDB.pm
bschmalhofer: That is interesting. The change was triggered by a failure in the test suite. Looks like both variants work, perhaps accidentally.

I'll fix this to the documented variant with ';' between the trailing attributes.

Kernel/System/MailQueue.pm
bschmalhofer: Fine with, I'll move that line back.

Kernel/System/DB.pm
svenoe: Let's keep the mysql for now, I think. If everything runs for some time, and we are happy with the changes, we can maybe just rename it in a second step, I guess this is not really relevant to the content and then the individual changes will be smaller.

Kernel/System/DB.pm
bschmalhofer: Staring at it again I concluded that the only intent of 	
```
my $rows = $sth->rows;
($rows == 0) ? "0E0" : $rows;
```
is to allow code like `if ($sth->do(...)) {...}`. So I think it is safe to simply check for definednes in `Kernel::System::DB::Do()`

bschmalhofer:
The test suit is still succeeding after the recent changes. Merging now.

PR #5070, opened 2026-01-22
Implement native support for RHEL 9.7

Splayfery:
This PR includes all configuration files and changes to support a native installation of OTOBO 11.x. Tested on RHEL 9.7 but might also work in different versions.

The following changes have been made:
- Create nginx config files for ssl and non-ssl reverse proxy support
- Include `nginx` user group in `bin/otobo.SetPermissions.pl`
- Create `cpanfile.plackup` for all required perl packages on RHEL 9.7 (removed optional options for gazelle and mysql) 
- Create `scripts/build-rhel-modules.sh` to automate the deployment of the required perl packages to a tar.gz file
- Create systemunit files for `otobo-web` and `otobo-daemon` (The service will be controlled using systemctl)

The build script has also been tested on the `redhat/ubi9:9.7` docker image to provide the opportunity for a deployment using Github Actions or Gitlab CI.

A complete documentation of the installation process has been created under https://github.com/RotherOSS/doc-otobo-installation/tree/issue-%23151-rhel.

### DIFF
--- a/scripts/build-rhel-modules.sh
+++ b/scripts/build-rhel-modules.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+# --
+# OTOBO is a web-based ticketing system for service organisations.
+# --
+# Copyright (C) 2001-2020 OTRS AG, https://otrs.com/
+# Copyright (C) 2019-2026 Rother OSS GmbH, https://otobo.io/
+# --
+# This program is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# --
+
+set -Euo pipefail
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+PROJECT_ROOT="$SCRIPT_DIR/.."
+TARGET_LIB="$PROJECT_ROOT/install/local"
+ARCHIVE_NAME="otobo-deps-11.0-rhel-9.7.tar.gz"
+# Directory at /opt/otobo to compress to tar.gz
+ARCHIVE_DIR="install"
+
+if ! command -v cpanm &> /dev/null; then
+    echo "cpanm not found. Installing using dnf..."
+    
+    if [ "$EUID" -ne 0 ]; then
+        echo "Installatin failed. Please provide sudo credentials."
+        sudo dnf install -y perl-App-cpanminus
+    else
+        dnf install -y perl-App-cpanminus
+    fi
+else
+    echo "cpanm is already present. Skipping Installation."
+fi
+
+cd "$PROJECT_ROOT"
+
+if [ -d "$TARGET_LIB" ]; then
+   echo "$TARGET_LIB already exists. Skipping."
+else
+   echo "Creating $TARGET_LIB ..."
+   mkdir -p "$TARGET_LIB/lib/perl5"
+fi
+
+PERL5LIB="$PROJECT_ROOT/install/local/lib/perl5"
+
+# Overwritte default cpanfile for cpanm build
+cp cpanfile.plackup cpanfile -f
+
+if [ -f "cpanfile" ]; then
+    echo "Installing dependencies to $TARGET_LIB..."
+    cpanm --local-lib "$TARGET_LIB" --self-contained --notest --installdeps .
+else
+    echo "Error: Couldn't find cpanfile in $PROJECT_ROOT!"
+    exit 1
+fi
+
+# Make binaries executable
+chmod +x ${TARGET_LIB}/bin/*
+
+tar -zcf ${ARCHIVE_NAME} -C . ${ARCHIVE_DIR}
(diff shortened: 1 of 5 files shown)


bin/otobo.SetPermissions.pl
```diff
@@ -38,7 +38,7 @@
 
 # for determining the group
 my %DefaultGroupNames = (
-    Apache => [qw(wwwrun apache www-data www _www)],
```
bschmalhofer, line 41: Is it sensible to give the OTOBO files to the group nginx? Nginx is just a proxy and has no need to access these files. My idea would be to detect the non-Apache use case and use the PSGI group list, effectively using the group 'otobo'. Unfortunately I have no good idea how to detect this case.

cpanfile.plackup
bschmalhofer, line 1: The goal is that no changes are necessary for this file. I propose to add auth:openidconnect in _bin/otobo.CheckModules.pl_ . The other changes are basically removing comments and commented out code. I propose to keep those, in order to stay in line with the Docker cpanfile. 

Plack::Hander::Gazelle is part of Gazelle, so it doesn't need to be specified here.

scripts/build-rhel-modules.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -Euo pipefail
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+PROJECT_ROOT="$SCRIPT_DIR/.."
+TARGET_LIB="$PROJECT_ROOT/install/local"
+ARCHIVE_NAME="otobo-deps-11.0-rhel-9.7.tar.gz"
```
bschmalhofer, line 23: In future this name could be determined from the environment.

cpanfile.plackup
```diff
@@ -112,17 +112,13 @@ requires 'Plack::Middleware::ReverseProxy';
 # Feature 'apache:mod_perl' is not needed
 
 
-# Feature 'auth:openidconnect' is not needed
+requires 'auth:openidconnect';
```
bschmalhofer: This should not work. Is Crypt::JWT needed? If so I'll update _bin/otobo.CheckModules.pl_ .

var/nginx/nginx-ssl.conf
bschmalhofer, line 1: The nginx config files should be in a subdirectory, maybe below 'var'. That the Apache files are located here is likely a quirk, resulting from the installation layout of Apache.

scripts/build-rhel-modules.sh
bschmalhofer, line 1: I think that the convention is to use underbars in file names.

var/systemd/otobo-web.service
```diff
@@ -0,0 +1,18 @@
...
+
+[Service]
+Type = simple
+User = otobo
+Group = otobo
+WorkingDirectory = /opt/otobo
+Environment = OTOBO_HOME=/opt/otobo
+Environment = PERL5LIB=/opt/otobo/install/local/lib/perl5
+Environment = PATH=/opt/otobo/local/bin:/opt/otobo/install/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+ExecStart = /opt/otobo/bin/docker/entrypoint.sh web
```
bschmalhofer, line 14: This is fine, even if a name containing 'docker' is a bit surprising here.

scripts/build-rhel-modules.sh
Splayfery: All scripts in this directory are using a "-" symbol to split words. Should this still be changed?

bschmalhofer: Ah, sorry I didn't see this. Then "-" is fine I guess.

cpanfile.plackup
```diff
@@ -111,21 +111,14 @@ requires 'Plack::Middleware::ReverseProxy';
 
 # Feature 'apache:mod_perl' is not needed
 
+# Required for authentication via OpenIDConnect.
```
bschmalhofer: I still recommend to make no changes in _cpanfile.plackup_ as it is a generated file. When changes are needed then make the change _bin/otobo.CheckModules.pl_ in regenerate _cpanfile.plackup_ .

scripts/build-rhel-modules.sh
bschmalhofer: Resolving the conversation as the above was just a note for the future.

scripts/build-rhel-modules.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+PERL5LIB="$PROJECT_ROOT/install/local/lib/perl5"
+
+# Overwritte default cpanfile for cpanm build
+cp cpanfile.plackup cpanfile -f
+
+if [ -f "cpanfile" ]; then
+    echo "Installing dependencies to $TARGET_LIB..."
+    cpanm --local-lib "$TARGET_LIB" --with-feature "gazelle" --with-feature "db:mysql" --with-feature "auth:openidconnect" --self-contained --notest --installdeps .
```
bschmalhofer: Switching to carton would be nice here. Carton supports version pinning via cpanfile.docker.snapshot.

bschmalhofer: However, this is a topic for the next iteration. Let's keep it as it currently is.

bin/otobo.SetPermissions.pl
bschmalhofer: Is there an Idea how to detect that we are actually in the PSGI case? The only idea I have is to check whether Apache2 is installed.

scripts/build-rhel-modules.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
bschmalhofer, line 1: @Splayfery , could you add the copyright blurb, just like in the other scripts

bschmalhofer: Thanks.

stefanhaerter:

Issue #4944, opened 2025-12-15
Update jquery-ui to version 1.14.1


chzauleck:

chzauleck:
This should be looked at again, there are some icon changes that we might not want, possibly related to some theme mistake?

PR #5072, opened 2026-01-22
Issue #4944 update jquery UI to 1.14.1

chzauleck:


stefanhaerter:

Issue #5084, opened 2026-01-27
Performance improvement of Kernel::System::HTMLUtils::ToAscii()


bschmalhofer:
This issue was triggered by the pull request #5079 by @LordAro . Here is the original description:

> Our OTOBO instance ran into a problem where the MailFetch cron was just timing out every time
> 
> Use of perl bin/otobo.Console.pl Maint::PostMaster::MailAccountFetch --force-pid --debug and many print statements later (I'm sure there must be a better way of debugging, but I don't know it), I narrowed the issue down to the blockquote processing section of Kernel::System::HTMLUtils::ToAscii
> 
> I figured the .*? capture group looked suspicious, so swapped it out with [^>]* and suddenly everything processed quickly. All fixed, time for a PR.
> 
> (I'm sure I was told not to use regex to parse HTML [s̴̰͊ö̵̺́m̵̝̂e̵̢̛w̶͉̒h̶̞̎e̵͓̔r̵̟͐e̸͈̕](https://stackoverflow.com/a/1732454/995325) ;-))
> 
> I don't know exactly what was in the email body to cause such an issue (the email itself was "only" 700kB in size including attachments), but this seems better anyway
> 
> (shout at me if you'd like a corresponding issue)

The PR looks fine, but warrants a bit of inspection.  Also, I'll add at least a couple of test cases. That parsing HTML with regexes can lead to surprising results is understood.

bschmalhofer:
I decided to not include the changes in 11.0.15 as there are still a couple of TODOs:

TODO:

- [x] revert the removal of the capture group
- [x] just for learning: try the effect of using a named capture
- [x] add test case with a broken blockquote start tag
- [x] create issue for the case with two blockquotes

bschmalhofer:
A funny effect occurs when there are two blockquote elements in the input. Only the lines o f the first blockquote are prepended with '> '.  This behavior is exemlified in this test case in _ToAscii.t_:

>     {
>         # note that the '> ' is added only for the first blockquote element
>         Input  => qq{before<blockquote>line A<br>line B</blockquote>after<br/>before<blockquote>line C<br>line D<blockquote>},
>         Result => qq{before> line A\n> line B\nafter\nbeforeline C\nline D},
>         Name   => 'ToAscii - two <blockquote> elements'
>     },
>

bschmalhofer:
[...]

PR #5099, opened 2026-01-29
Issue #5084 to ascii blockquote

bschmalhofer:


Kernel/System/HTMLUtils.pm
```diff
@@ -135,17 +135,22 @@ sub ToAscii {
     my %Cite;
     $Counter = 0;
     $Param{String} =~ s{
-        <blockquote(.*?)>(.+?)</blockquote>
+        <blockquote [^>]*>(.+?)</blockquote>
```
stefanhaerter: Is the whitespace before the square bracket on purpose?

Kernel/System/HTMLUtils.pm
```diff
@@ -135,17 +135,22 @@ sub ToAscii {
...
     {
-        my $Ascii = $Self->ToAscii(
-            String => $2,
-        );
-        # force line breaking
+        # ToAscii() uses regexes internally. Apparently this resets some regex global counter,
+        # which has the effect that the replacement is done only for the first <blockquote>.
+        my $Ascii = $Self->ToAscii( String => $1 );
+
+        # force line breaking, note that below there is a different $1 than above
```
stefanhaerter, line 145: Would the use of named captures improve readability regarding the `$1` topic?

stefanhaerter (approved):
LGTM 👍🏻

Kernel/System/HTMLUtils.pm
bschmalhofer: Yes, I included it for readability only. The substitution does have the 'x' modifier.

But I think that I will add the capturing parentheses again. The captured value is not used, but I have to draw the line somewhere before making too many changes.

stefanhaerter: Ah, I was suspecting this and looked for the modifier, but didn't find it at first. Thank you :)

Kernel/System/HTMLUtils.pm
bschmalhofer: I will create a new issue that addresses the problem with two blockquotes.

And yes, named captures would improve readablity there. But I think that named captures would not avoid the problem with two blockquotes, though I'm not really sure.

stefanhaerter: Ah, I didn't think of named captures as a way to solve the actual problem - I just stumbled on your comment

> note that below there is a different $1 than above

And to me this sounded like a good opportunity to enhance readability.

bschmalhofer: I switched to using named captures. Not sure whether that makes the regex more readable.

bschmalhofer:
Let's merge this only after the release of OTOBO 11.0.15.

PR #5123, opened 2026-02-05
Remove unused perl modules & optimize build script

Splayfery:
- Remove unused perl packages from CheckModules.pl due to compatibilitz issues
- Explicitly require cpanm default mirror for building perl modules. This makes the build process clearly faster

### DIFF
--- a/scripts/tools/build-rhel-modules.sh
+++ b/scripts/tools/build-rhel-modules.sh
@@ -53,7 +53,7 @@ cp cpanfile.plackup cpanfile -f
 
 if [ -f "cpanfile" ]; then
     echo "Installing dependencies to $TARGET_LIB..."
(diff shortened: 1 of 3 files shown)


scripts/tools/build-rhel-modules.sh
```diff
@@ -53,7 +53,7 @@ cp cpanfile.plackup cpanfile -f
 
 if [ -f "cpanfile" ]; then
     echo "Installing dependencies to $TARGET_LIB..."
-    cpanm --local-lib "$TARGET_LIB" --self-contained --notest --installdeps .
+    cpanm --local-lib "$TARGET_LIB" --self-contained --notest --mirror "http://www.cpan.org" --mirror-only --installdeps .
```
bschmalhofer: Could you test whether it also works with https://www.cpan.org ?
Out of curiousity, was in the  old version a different mirror used?

PR #5175, opened 2026-02-20
Issue #5176: Cleaned up two duplicates in article event config.

### LINKED ISSUE #5176: Duplicates in article event system configuration

stefanhaerter:
closes #5176


bschmalhofer (commented):
Removing the duplicates looks sensible to me. As far as I see the settings are used in the config for notifications and the generic agent. Are the duplicates visible there? If so the change should be part of an issue. If not I'll merge the PR.

Issue #1788, opened 2022-05-19
Allow User Modification of SysConfig settings


bschmalhofer:
Looks like allowing user modification of sysconfig settings is not really working. For a started, the Subaction UserModified seems to be not implemented in _Kernel/Modules/AdminSystemConfiguration.pm_ .

_Originally posted by @bschmalhofer in https://github.com/RotherOSS/otobo/issues/1776#issuecomment-1131615469_

svenoe:
https://ideas.otobo.io/posts/5

PR #5218, opened 2026-03-02
Add advanced settings to agent personal perferences

urbalazs:
**New feature:** advanced settings in agent personal preferences.

This feature allows agents to change more settings in their personal preferences. A new section called "Advanced" is displayed where agents can fine tune some settings related to agent interface.

The logic of this feature came from ((OTRS)) Community Edition, but the switch was part of Business Solution package. This means that an expert agent already had the possibility to activate this feature in the system, but it was not activated by default - only for Business Solution customers.

Since this can be considered as "new feature", I create this pull request against `rel-11_1` only. The logic is also part of the previous versions, so feel free to back-port this feature to `rel-11_0` and `rel-10_1` if you think, it would be useful for current customers. This change is backward compatible.

Thanks to [László Gyaraki](https://github.com/gyarakilaszlo) to find the way how to activate this feature.

### DIFF
--- a/Kernel/Config/Files/XML/Framework.xml
+++ b/Kernel/Config/Files/XML/Framework.xml
@@ -3712,7 +3712,7 @@ You can log in via the following URL:
             <Item ValueType="String" ValueRegex="">http://host.example.com/cgi-bin/login.pl</Item>
         </Value>
     </Setting>
-    <Setting Name="Frontend::CustomerUser::Item###1-GoogleMaps" UserPreferencesGroup="Advanced" UserModificationPossible="1" Required="0" Valid="1">
+    <Setting Name="Frontend::CustomerUser::Item###1-GoogleMaps" Required="0" Valid="1">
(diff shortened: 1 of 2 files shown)


stefanhaerter:
Related to https://github.com/RotherOSS/otobo/issues/1788

stefanhaerter:
I just started testing the PR. Unfortunately, the module AgentPreferences produces the follwoing browser console error when opening a modifiable sysconfig setting:

image

I'll start looking into it.

stefanhaerter:
Firstly, I added `Core.Agent.Admin.SystemConfiguration.js` and its dependency `thirdparty/clipboardjs-1.7.1/clipboard.min.js` to the sysconfig `Loader::Module::AgentPreferences###001-Framework`.

This however led to the situation that the user sees all sysconfig settings in the tree in the left sidebar, regardless if she or he can modify them or not.

I think the core issue is around the ResetSetting dialogue. `Core.SystemConfiguration.js` tries to take a function from `Core.Agent.Admin.SystemConfiguration.js` for this purpose. Maybe the function can just be copied over.

PR #5234, opened 2026-03-06
Issue #5233: Prevent uninitialized value warning.

### LINKED ISSUE #5233: Further stabilize GetFieldStates unit test
At least on rel-11_1, the ACL system module complained about use of uninitialized value, which seemingly is due to the fact that not the full ACL data is passed on from the test.

Also, this seemingly prevents setting the ACL to invalid correctly. Therefor, it should be tested if the ACL actually is invalid after updating.

Based on #3813.

stefanhaerter:
Also added testing if ACLs actually are invalid after updating and took care of two unused variables.

closes #5233


MichaelThumes (approved):
looks good, let's give it a try

Issue #5148, opened 2026-02-11
Include package ImportExportStandardObjects into core


stefanhaerter:

stefanhaerter:
Remaining ToDos are:

- [x] write documentation
- [x] add unit tests with simple test cases for each object type
- [x] include adaptions for service and SLA

stefanhaerter:
Documentation has been written and merged, closing the issue.

PR #5246, opened 2026-03-12
Issue #5148: Fixed valid column for generic agent export screen.

stefanhaerter:
Included fix applied to package.

references #5148


bschmalhofer:
LGTM

PR #5297, opened 2026-03-25
Issue #5296: Return array in each case of SearchObjects.

### LINKED ISSUE #5296: Unify return value of sub SearchObjects across reference dynamic field drivers

stefanhaerter:
closes #5296


MichaelThumes (approved):
looks sane to me

PR #5299, opened 2026-03-25
Issue #5298: Return if customer user to object id does not exist.

### LINKED ISSUE #5298: Catch nonexistent customer user identifier in reference driver sub SearchObjects

stefanhaerter:
closes #5298


MichaelThumes (approved):
looks like advertised

Issue #5078, opened 2026-01-26
Duplicate CustomerNavigationBar in CustomerTicketMessage.pm on validation error


kuwadigital:
### Environment                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                 
  - Server OS Distribution: Ubuntu 22.04 LTS                                                                                                                                                                                                                                     
  - OTOBO installation type: Docker                                                                                                                                                                                                                                              
  - Browser: All browsers (Chrome, Firefox, Safari)                                                                                                                                                                                                                              
  - OTOBO version: 11.0.x                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                 
  ### Expected behavior                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                 
  When a customer submits the new ticket form (`CustomerTicketMessage`) with validation errors, the customer navigation bar should be displayed only once and the expand button schould work correctly.                                                                                                                       
                                                                                                                                                                                                                                                                                 
  ### Actual behavior                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                 
  The customer navigation bar is rendered twice on the page when the form submission fails due to validation errors (e.g., missing required fields).                                                                                                                             
                                                                                                                                                                                                                                                                                 
  ### How to reproduce                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                 
   
  0. In the Agent Interface, deactivate all  Ticket-Fields (Queues, Services ...) and keep only the Subject and RichText                                                                                                                                                                                                                                                                     
  1. Go to the Customer Interface                                                                                                                                                                                                                                                
  2. Click on "New Ticket" (CustomerTicketMessage)                                                                                                                                                                                                                               
  3. Leave required fields empty (e.g., Subject or Body)                                                                                                                                                                                                                         
  4. Click "Submit"                                                                                                                                                                                                                                                              
  5. Observe the page with validation errors                                                                                                                                                                                                                                     
  6. The navigation bar appears twice on the page  (can be find out by analysing in the console of the navigator)
  7. The Navigation is not reacting anymore when we the user try to expand the NavigationBar                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                 
  ### Additional information                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                 
[...]

PR #5271, opened 2026-03-23
Issue #5078: Fixed duplicate CustomerNavigationBar and gray overlay on error

rmfalves:

PR #5302, opened 2026-03-25
Issue #5078: Fixed duplicate CustomerNavigationBar and gray overlay on error

rmfalves:


stefanhaerter (changes requested):
This has rel-11_0 as target, just like #5271. I suggest to change the target to rel-10_1.

rmfalves:
Invalid PR: The target branch is incorrect, it should be `rel-10_0`. Closing.

PR #5306, opened 2026-03-26
Issue #5009: Position error dialogue relative to viewport.

### LINKED ISSUE #5009: Submitting an erroneous postmaster filter leads to being stuck with overlay
See this video:

https://github.com/user-attachments/assets/32cab027-2520-429d-a828-872b89eb06cd

The effect does also occur when something is selected in the dropdown upon error and it does occur on creation as well as on editing.

Also observed on rel-11_1, so my guess is that all branches from rel-10_1 upwards are affected.

stefanhaerter:
This is achieved by setting position to fixed, which causes the top value to be calculated relative to the viewport.

closes #5009


MichaelThumes (approved):
@chzauleck approved, looks as advertised, let's go with it

PR #5310, opened 2026-03-27
Tidied.

stefanhaerter:
Removed outcommented duplicated keys from password configs.


bschmalhofer:
Not sure about this, in a way it makes sense to have an example for Password Regexes somewhere. But _Kernel/Config/Defaults.pm_ might be a better place for that.

stefanhaerter:
@svenoe agreed to leaving it in. Closing the PR.

PR #5321, opened 2026-03-30
Issue #5009: Fixed modal dialog in admin process management.

### LINKED ISSUE #5009: Submitting an erroneous postmaster filter leads to being stuck with overlay

stefanhaerter:
closes #5009


stefanhaerter:
Closing because actual root cause in in JS and should be fixed there.

Issue #5334, opened 2026-04-01
Do not use the double period ".." in the Perl module search path


bschmalhofer:
> @MichaelThumes pointed out that components in `@INC` use the parent dir ".." syntax. Let's change that for rel-11_1.
> 
> ```
> $INC_array = [
>                '/opt/otobo/bin/psgi-bin/../../Custom',
>                '/opt/otobo/bin/psgi-bin/../../Kernel/cpan-lib',
>                '/opt/otobo/bin/psgi-bin/../..',
>                '/opt/otobo/local/lib/perl5/5.40.3/x86_64-linux-gnu',
>                '/opt/otobo/local/lib/perl5/5.40.3',
>                '/opt/otobo/local/lib/perl5/x86_64-linux-gnu',
>                '/opt/otobo/local/lib/perl5',
>                '/opt/otobo_install/local/lib/perl5/5.40.3/x86_64-linux-gnu',
>                '/opt/otobo_install/local/lib/perl5/5.40.3',
>                '/opt/otobo_install/local/lib/perl5/x86_64-linux-gnu',
>                '/opt/otobo_install/local/lib/perl5',
>                '/usr/local/lib/perl5/site_perl/5.40.3/x86_64-linux-gnu',
>                '/usr/local/lib/perl5/site_perl/5.40.3',
>                '/usr/local/lib/perl5/vendor_perl/5.40.3/x86_64-linux-gnu',
>                '/usr/local/lib/perl5/vendor_perl/5.40.3',
>                '/usr/local/lib/perl5/5.40.3/x86_64-linux-gnu',
>                '/usr/local/lib/perl5/5.40.3'
>              ];
> ``` 

 _Originally posted by @bschmalhofer in [#5332](https://github.com/RotherOSS/otobo/issues/5332#issuecomment-4169084517)_

bschmalhofer:
Only _bin/psgi-bin/otobo.psgi_ will be considered here. Most other scripts already avoid the double period.

bschmalhofer:
PR is merged. The OTOBO web app still works. Closing this issue.

PR #5337, opened 2026-04-01
Issue #5334: avoid .. in @INC

bschmalhofer:
This makes the actually used directories more obvious


bschmalhofer:
Ignoring the complaints about comment lines.

PR #5395, opened 2026-04-15
Tidied.

stefanhaerter:
Removed unused variable.


scripts/test/Loader.t
```diff
@@ -47,7 +47,6 @@ $Kernel::OM->ObjectParamAdd(
         RestoreDatabase => 1,
     },
 );
-my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
```
bschmalhofer, line 50: The variable is not needed, but calling the constructor of `Kernel::System::UnitTest::Helper` is necessary for triggering the database restore after completion of the script.

Edit: actually I'm don't see why a database cleanup is needed in the script. So lines 45-49 can be removed as well.

stefanhaerter: Adjusted the PR accordingly.

PR #5411, opened 2026-04-20
Issue #5358: update unitest to new substitution rule. also test corre…

MichaelThumes:
…ct vector.


MichaelThumes:
merging with ok from @svenoe

PR #5432, opened 2026-04-22
Tidied.

stefanhaerter:
Added no critic marker to make clear that this violation is on purpose.


stefanhaerter:
Closing as the upstream policy does complain - no clue yet why.

Issue #5350, opened 2026-04-07
Integrate the Elasticsearch extension package into the core


rmfalves:
The Elasticsearch functionalities should be integrated into the core and should be fully operational when the FAQ package is installed.

stefanhaerter:
Edit: Changed the title to match package name to be able to find the issue via search.

PR #5433, opened 2026-04-22
Issue #5350: Integrate elasticsearch extended

rmfalves:


MichaelThumes (changes requested):
pls check
- ElasticsearchQuickResult.pm if Subaction "ViewDocs" ist still needed at all. If  not please remove the whole Subaction
- I got sporadic warnings that look like this:
```
Use of uninitialized value $ESSearchQuery in substitution (s///) at /opt/otobo/Kernel/Modules/AgentElasticsearch.pm line 57.
Use of uninitialized value $ESSearchQuery in substitution (s///) at /opt/otobo/Kernel/Modules/AgentElasticsearch.pm line 58.
```
I think that might just be related to 'FulltextES' Query param to be empty (undef), please have a look if this can be avoided like with a 

```Perl
my $ESSearchQuery = $ParamObject->GetParam( Param => 'FulltextES' ) // '';
```

or similar.

MichaelThumes (approved):
this looks good to me

Issue #5436, opened 2026-04-23
Integrate the ProcessTicketTemplate package into the core


rmfalves:
The ProcessTicketTemplate functionalities should be integrated into the core and should be fully operational as standard features.

stefanhaerter:
PR merged. Set milestone and labels, closing the issue.

stefanhaerter:
Corrected the package name in title to enable searching for it.

PR #5453, opened 2026-04-28
Issue #5436 integrate process ticket templates

rmfalves:


MichaelThumes (approved):
had a clarifying chat with @rmfalves about event binding in Core.Agent.TicketProzess.js - assuring the second on-change bind after AJAX update is necessary and does not lead to a double bind.

MichaelThumes:
ignoring ESLint related errors from codepolicy

PR #5470, opened 2026-04-30
Issue #5469: Harden group filter against empty strings.

### LINKED ISSUE #5469: DF Reference Agent doesn't work without filtering for groups
You can't select an Agent in a Dynamic Field of type Reference Agent without selecting a group in the Agent Group setting of the Dynamic Field configuration.

Configuration without a group:

Image

Result:

Image

Configuration with a group:

Image

Result:

Image

I as an administrator would expect the behaviour to be consistent: If you have a filter option, you filter, if you have no filter option, you don't filter and see everything.

Tested with OTOBO 11.0.15

stefanhaerter:
closes #5469


stefanhaerter:
Closed in favor of #5477

Issue #5521, opened 2026-05-18
Performance improvements for rhel build script


Splayfery:
Currently the `scripts/tools/build-rhel-modules.sh` script has a relatively poor performance when executed by an CI-Runner.

After testing this is mainly caused by the following reasons:
- The script only relies on one mirror
- The `cpan.org` mirror is used which provides slower download speeds compared `cpan.metacpan.org`
- The `cpan.org` mirror doesn't archive some older pinned versions of perl modules

Affected build command:
```bash
cpanm --local-lib "$TARGET_LIB" --self-contained --notest --mirror "http://www.cpan.org" --mirror-only --installdeps .
```

PR #5526, opened 2026-05-19
Update cpanminus build command

Splayfery:
Refering to #5521.

> Currently the `scripts/tools/build-rhel-modules.sh` script has a relatively poor performance when executed by an CI-Runner.
> 
> After testing this is mainly caused by the following reasons:
> - The script only relies on one mirror
> - The `cpan.org` mirror is used which provides slower download speeds compared `cpan.metacpan.org`
> - The `cpan.org` mirror doesn't archive some older pinned versions of perl modules

The following changes have been tested on an internal gitlab ci runner intance.


bschmalhofer (approved):
Not passing `--mirror-only` makes cpanm use the Metacpan API for resolving dependencies. This looks fine to me, as it is the default behavior of cpanm.

Issue #5632, opened 2026-06-10
Check for 32-bit Perl in Dev::Code::CPANAudit


bschmalhofer:
Currently `CPAN::Audit` alerts about https://nvd.nist.gov/vuln/detail/CVE-2026-8376 . However, this CVE is only about 32bit Perl installations. I think that 32bit Perls are a very unusual setup for web application servers.

TODO:

- [x] Check `$Config{use64bitall}` and set `is_relevant_for_otobo = 0` in that case.

bschmalhofer:
Done, closing this issue.

PR #5633, opened 2026-06-10
Issue #5632: exempt CPANSA-perl-2026-8376 on 64bit Perls

bschmalhofer:


bschmalhofer:
Ignore the compile check that fails because of `use Config;`. See https://github.com/RotherOSS/codepolicy/issues/144

Issue #4258, opened 2025-04-07
Include package ExtendedArticleEdit into core


stefanhaerter:
See title.

stefanhaerter:
Current question is: How to design the migration script such that the system behaves the same before and after migration in each case (package installed vs. package not installed previously)?

stefanhaerter:
PR merged, closing the issue.

Issue #3861, opened 2024-10-03
Clean up Ticket Mask Styles


chzauleck:
cleaning up html structure and css styling of forms, especially the ones which can include dynamic fields to make styling and future improvements easier and more consistent

PR #5635, opened 2026-06-10
Issue #3861: Adapted article edit template.

stefanhaerter:
Also related to #4258.

references #3861


MichaelThumes (approved):
Looks good to me as well

PR #5636, opened 2026-06-10
Tidied.

stefanhaerter:
Replaced 11.0 with 11.1.

### DIFF
--- a/scripts/DBUpdateTo11_1/Base.pm
+++ b/scripts/DBUpdateTo11_1/Base.pm
@@ -50,7 +50,7 @@ This modules provides methods for the task modules that are located in F<scripts
 
 =head2 new()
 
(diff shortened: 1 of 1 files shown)


scripts/DBUpdateTo11_1/Base.pm
```diff
@@ -50,7 +50,7 @@ This modules provides methods for the task modules that are located in F<scripts
 
 =head2 new()
 
-The constructor provided by the base module and used in F<scripts/DBUpdateTo11.0.pm>.
+The constructor provided by the base module and used in F<scripts/DBUpdateTo11.1.pm>.
```
bschmalhofer: Could you change it to "is provided" and "is used" ?

Issue #5639, opened 2026-06-10
Inconvenient name for rhel tar archive


Splayfery:
Since `scripts/tools/build-rhel-modules.sh` is used for other OTOBO versions in the future it is very misleading to define the static tar archive name `ARCHIVE_NAME="otobo-deps-11.0-rhel-9.7.tar.gz`.

For this a more generic name should be considered.

PR #5640, opened 2026-06-10
Make rhel archive name more generic

Splayfery:
Referring to #5639.

The tar archive name created by the `scripts/tools/build-rhel-modules.sh` script has been changed from `otobo-deps-11.0-rhel-9.7.tar.gz` to `otobo-deps-rhel.tar.gz` to make it more generic for further OTOBO versions and RHEL releases.

Because the script relies on the cpanfile.plackup from the cloned repository locally, it's functionality is independent of the current OTOBO version because it always uses the local plackup file which already gets updated on new OTOBO versions if necessary.

The correct archive names for release packages on https://ftp.otobo.org are getting handeled by a dedicated CI-Runner.


Splayfery:
@bschmalhofer Please make sure to also push these changes to rel-11_1 because the new CI-Runner-Configuration now also builds packages from this branch

Issue #5665, opened 2026-06-12
Use more common database column size for the new  OIDC tables


bschmalhofer:
This was noticed while working #5659. Installation via _quick_setuo.pl_ and via the installer failed because of an SQL error. The error was about unique indexes on TEXT fields. It looks like MySQL is in that regards more picky than MariaDB. The fix was to change the column size in the new table from 256 to 255.
```
DBD::MariaDB::st execute failed: BLOB/TEXT column 'name' used in key specification without a key length at /opt/otobo/bin/docker/../../Kernel/System/DB.pm line 720.
ERROR: quick_setup-10 Perl: 5.40.4 OS: linux Time: Fri Jun 12 13:52:36 2026

 Message: BLOB/TEXT column 'name' used in key specification without a key length, SQL: 'CREATE TABLE oidc_profiles (
    id INTEGER NOT NULL AUTO_INCREMENT,
    name TEXT NOT NULL,
    client_id TEXT NOT NULL,
    client_secret TEXT NOT NULL,
    redirect_uri TEXT NOT NULL,
    openid_config TEXT NOT NULL,
    ssl_options TEXT NOT NULL,
    misc TEXT NOT NULL,
    ttl INTEGER NOT NULL,
    create_time DATETIME NOT NULL,
    create_by INTEGER NOT NULL,
    change_time DATETIME NOT NULL,
    change_by INTEGER NOT NULL,
    valid_id SMALLINT NOT NULL,
    PRIMARY KEY(id),
    UNIQUE INDEX oidc_profiles_name (name)
)'

 Traceback (28): 
   Module: main::ExecuteSQL Line: 656
   Module: main::Main Line: 277
   Module: bin/docker/quick_setup.pl Line: 1073

BLOB/TEXT column 'name' used in key specification without a key length
```

bschmalhofer:
Actually this is a bit more involved. https://doc.otobo.org/manual/installation/11.0/en/content/requirements.html states that MySQL 8,4 is required. This is quite sensible as indicated by https://endoflife.date/mysql .

MySQL 5.6 hat an index size limit of 767 octets. That is why many columns have a size of `int( 767 / 4 ) = 191`  UTF-8 characters. According to https://dev.mysql.com/doc/refman/8.4/en/create-index.html they regular max key length for InnoDB in MySQL 8.4 is  3072 octets. Thus a `VARCHAR(255)` column can be indexed just fine.

There are two unlikely cases:
[...]

PR #5667, opened 2026-06-12
Issue #5665: use the more common column size 255

bschmalhofer:
instead of the size 256. This means that the column are not created as TEXT fields. Thus MySQL can create unique indexes for these columns.


MichaelThumes (changes requested):
basically yes looks good to me.

but pls check https://github.com/RotherOSS/otobo/blob/rel-11_1/scripts/DBUpdateTo11_1/DBUpdateOIDC.pm which should be aligned?

bschmalhofer:
I also adapted the upgrade script.

PR #5670, opened 2026-06-15
Issue #5494: Doubled the name length for ACLs in HTML.

### LINKED ISSUE #5494: Longer name for ACLs
The ACL names are always quite restricted length-wise. If you try to build a structured ACL concept, e.g. with the affected action, the used process, the use case and what is happening, the names can become quite long. Due to the restricted length of the name, this can be very challenging and you have to work with abbreviations, which makes it confusing and not well-readable when you are in the ACL overview and try to gain an overview over the concept.
Even more annoying, you copy an ACL, it creates the copy with a name suffix e.g. "(Copy) 1", which is longer than the limit. When you now try to re-number your numbered prefix, you delete some digits. Now you can't add new digits (e.g. 200 - 101 => 200 - 102), because with the suffix it is too long for the field. Now you have to jump back in the field, delete the suffix, jump back to the prefix and now you can add your numbers. As it happens a lot (!) that you dance around the limit, you always stumble upon those issues.
Please enhance the field to allow more characters.

(It would also be great for the comment and for the description)

stefanhaerter:
It remains unclear why the HTML maxlength was 70 while the respective DB column allowed 191.

closes #5494

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminACLEdit.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminACLEdit.tt
@@ -87,7 +87,7 @@
                     <fieldset class="TableLike">
                         <label class="Mandatory" for="Name"><span class="Marker">*</span> [% Translate("Name") | html %]:</label>
                         <div class="Field">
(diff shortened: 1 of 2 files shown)


MichaelThumes (approved):
lgtm

Issue #5686, opened 2026-06-17
Possibly renamed / invalid Queues in System Configuration Setting


stefanhaerter:
Issue #3561 introduced value entity types for some queue system configuration settings, including default values. However, those default value queues might not exist, be invalid or have been renamed on a system.

Suggestion is to just omit the default values.

stefanhaerter:
Settings with known issues:

- `Elasticsearch::ExcludedQueues`
- `DashboardEventsTicketCalendar###Queues`

Settings with suspected issues:

- `EscalationSuspendStates`

stefanhaerter:
`EscalationSuspendStates` shouldn't pose a problem as the setting is invalid per default.

PR merged, closing the issue.

PR #5687, opened 2026-06-17
Issue #5686: Removed default values for some settings.

stefanhaerter:
Deployment may fail if queues are invalid, renamed or deleted and thus the setting values can't be validated.

references #5686


MichaelThumes (approved):
reviewed

Issue #5703, opened 2026-06-18
DynamicField Backend: Mitigate internal server error for missing dynamic field driver modules


stefanhaerter:

stefanhaerter:
PR merged, closing the issue.

PR #5707, opened 2026-06-19
Issue #5703: Revert invalidating dynamic fields without backend.

stefanhaerter:
This could probably cause inconvenience during migrations and updates.

references #5703


MichaelThumes (approved):
as discussed with @stefanhaerter , we keep the main fix that allows the migration to proceed, but paddle back in regards to trying to cleanup misconfigured or obsolete DFs at this points as it would cause pain further down the road.

Issue #5717, opened 2026-06-20
DBUpdates returns an error when ElasticSearch isn't configured


Splayfery:
### Environment

- Server OS Distribution: Ubuntu 24.04 LTS
- OTOBO installation type: Native Nginx
- Browser: --
- OTOBO version: 11.1.x

### Behavior

When running the `scripts/DBUpdate-to-11.1.pl` file on an native installation that doesn't have ElasticSearch configured, the script returns thef `Migration failed` output even though it successfully finished. This should not be the case.

``` bash
otobo@otobo-test:~$ /opt/otobo/scripts/DBUpdate-to-11.1.pl

Migration started ...
	Executing task 1/18 'Rebuild the configuration as files in Kernel/Config/Files/XML have changed' (RebuildConfig) ...
	Executing task 2/18 'Deactivate autoload modules in the SysConfig' (SysConfigDeactivateAutoloadModules) ...
	Executing task 3/18 'Increase translation content length' (DBUpdateTranslationLength) ...
	Executing task 4/18 'Add OIDC/OAuth2 database tables and fields.' (DBUpdateOIDC) ...
	Executing task 5/18 'Update mail_account table to support OIDC/OAuth2.' (DBUpdateOIDCMail) ...
	Executing task 6/18 'Add columns for namespace and process entity id to process element tables.' (DBAddProcessColumns) ...
	Executing task 7/18 'Add object_type column to acl table, with 'Ticket' as default value.' (DBAddACLObjectType) ...
	Executing task 8/18 'Update the list of installed packages' (UninstallMergedPackages) ...
	Executing task 9/18 'Add valid_id to postmaster_filter.' (DBUpdatePostMasterFilter) ...
	Executing task 10/18 'Add process dynamic fields to ticket zoom dynamic field screen configs.' (SysConfigUpdateTicketZoomDFScreens) ...
	Executing task 11/18 'Add X-OTOBO-From to PostmasterX-Header.' (SysConfigUpdatePostmasterXHeader) ...
	Executing task 12/18 'Migrate Article::Actions system configurations to new structure.' (SysConfigMigrateArticleActions) ...
	Executing task 13/18 'Update MultiValue attribute of Lens dynamic fields.' (UpdateLensDynamicFieldsMultiValue) ...
	Executing task 14/18 'Deactivates the system configuration setting for CustomerTicketSearch per default.' (SysConfigDeactivateCustomerTicketSearch) ...
[...]

PR #5720, opened 2026-06-21
Issue #5717: early exit when Elasticsearch is not active

bschmalhofer:
return successfully as an inactive Elasticsearch is a regular case

The relevant change is the a deployment is only done when the SysConfig was actually changed.


MichaelThumes (approved):
looks good to me

PR #5721, opened 2026-06-22
Issue #5703: Also prevent internal server error for extensions.

stefanhaerter:
references #5703


MichaelThumes (approved):
looks reasonable to me

Issue #5793, opened 2026-06-29
Rebuild config with the option --cleanup in do_update_tasks


bschmalhofer:
This issue concerns the upgrade from 11.0 to 11.1. Currently the sequence of commands in `do_update_tasks` in _bin/docker/entrypoint.sh_ is:       

1. Admin::Package::ReinstallAll --hide-deployment-info
2. Admin::Package::UpgradeAll
3. Maint::Config::Rebuild --deploy-acls --deploy-processes
       
The `Admin::Package::UpgradeAll` may do a SysConfig cleanup depending on whether there are packages installed and whether the packages have code upgrade tasks. See https://github.com/RotherOSS/otobo/blob/256d8aeaf95df0c624c31e257531cd0b6130fad6/Kernel/System/Package.pm#L5025 . I propose to explicitly do a cleanup in all cases, that pass `--cleanup` in the `Maint::Config::Rebuild` step. This resolves the ambiguity whether a cleanup was done. There might be a problem with upgrading packages that rely on that the cleanup hasn't been done yet. But that effect already happens with `Admin::Package::UpgradeAll`.

bschmalhofer:
The logic in `Admin::Package::UpgradeAll` does a config cleanup only when all packages could be upgraded smoothly. So a cleanup that does not respect this condition would remove configuration that is needed for manual rework.

PR #5794, opened 2026-06-29
Issue #5793: add --cleanup to Maint::Config::Rebuild

bschmalhofer:


svenoe:
We cannot do a forced cleanup here, it would wipe all SysConfigs of all packages which can only be upgraded manually.

bschmalhofer:
In this case there might be a problem lurking, as `Admin::Package::UpgradeAll` might already do a cleanup.

PR #5803, opened 2026-06-30
Issue #5801: skip pwd masking if pwd is empty

### LINKED ISSUE #5801: Password masking regex is too eager
### Environment

- Server OS Distribution: any
- OTOBO installation type: any
- Browser: any
- OTOBO version: all

### Expected behavior

Console/Command/Maint/PostMaster/MailAccountFetch should mask the password before writing to STDERR.  And only the password

### Actual behavior

The whole error msg is masked hiding not only the password, but the interesting bits of the error msg as well

### How to reproduce

run MailAccountFetch console command with --debug flag

### Additional information

the relevant code seems to be there since the OTRS fork:

https://github.com/RotherOSS/otobo/blob/rel-11_0/Kernel/System/Console/Command/Maint/PostMaster/MailAccountFetch.pm#L193

MichaelThumes:
In case of XOAUTH2/OAUTHBEARER pwd will be empty, which will make the regex match everything.

closes #5801

### DIFF
--- a/Kernel/System/Console/Command/Maint/PostMaster/MailAccountFetch.pm
+++ b/Kernel/System/Console/Command/Maint/PostMaster/MailAccountFetch.pm
@@ -193,7 +193,9 @@ sub Run {
             # Hide password contained in error message and print message back to standard error.
             # Please see bug#12829 for more information.
             if ($ErrorMessage) {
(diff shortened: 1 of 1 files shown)


stefanhaerter (approved):
Nothing to mask if there is no password. LGTM 👍🏻

PR #5819, opened 2026-07-02
Issue #5818: add perm check to sysconf ajax deploy

### LINKED ISSUE #5818: add permission check to sysconfig ajaxdeploy

MichaelThumes:
closes #5818

### DIFF
--- a/Kernel/Modules/AdminSystemConfigurationDeployment.pm
+++ b/Kernel/Modules/AdminSystemConfigurationDeployment.pm
@@ -325,7 +325,9 @@ sub _IsAdmin {
 
     # module permission check for action
     if (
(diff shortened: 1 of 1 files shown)


Kernel/Modules/AdminSystemConfigurationDeployment.pm
```diff
@@ -19,6 +19,8 @@ package Kernel::Modules::AdminSystemConfigurationDeployment;
 use strict;
 use warnings;
 
+use Kernel::System::VariableCheck qw(:all);
```
stefanhaerter, line 22: Why `qw(:all)`, if only `IsHashRefWithData` and `IsArrayRefWithData` are actually used?

(I know that it is done like that in quite some places, however I think it is not that good of a technique...)

Issue #5505, opened 2026-05-12
Remove title from AgentTicketResponsible


StefanAbel-OTOBO:
By default it doesn't make sense to have the Ticket title field in the Responsible dialog.

stefanhaerter:
PR merged, closing the issue.

PR #5821, opened 2026-07-02
Issue #5505: Also remove title from unit test.

stefanhaerter:
AgentTicketResponsible does not have a title field anymore by default. Easiest is to adapt the test.

references #5505


MichaelThumes (approved):
looks reasonable to me

PR #5848, opened 2026-07-10
Issue #5809: Added other masks as deactivated to config default.

### LINKED ISSUE #5809: Ticket::Frontend::RestorePendingInformation###Actions needs better defaults
### Environment

- OTOBO installation type: all
- Browser: all
- OTOBO version: 11.1.0-beta1

### Actual behavior

The default for ` Ticket::Frontend::RestorePendingInformation###Actions` is only active on `AgentTicketPending`

### Expected behavior

These are the masks, where the next state can be set by default:

`AgentTicketPending`, `AgentTicketCompose`, `AgentTicketEmailOutbound`, `AgentTicketPhoneOutbound`, `AgentTicketPhoneInbound`

They should all be configured by default.

Inho: Even better would be to activate this by default and make a blacklist here to exclude masks.

stefanhaerter:
closes #5809

### DIFF
--- a/Kernel/Config/Files/XML/Ticket.xml
+++ b/Kernel/Config/Files/XML/Ticket.xml
@@ -4303,6 +4303,10 @@
         <Value>
             <Hash>
                 <Item Key="AgentTicketPending">1</Item>
+                <Item Key="AgentTicketCompose">0</Item>
+                <Item Key="AgentTicketEmailOutbound">0</Item>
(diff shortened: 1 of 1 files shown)


73:
I do not want to be rude, but I am genuinely curious. Why is that not the default on all these masks? What is the reasoning behind this decision (as it is not obvious to me)?

Issue #5464, opened 2026-04-30
AgentTicketPhone: RichText editor looks strange if width and height are customized


stefanhaerter:
See this screenshot:

<img width="885" height="743" alt="Image" src="https://github.com/user-attachments/assets/9f17a3b0-b38b-4f79-8590-d43fd1f8ef1b" />

The richtext editor is overlapping with the attachment area.

Modified settings:

- `Ticket::Frontend::AgentTicketPhone###RichTextHeight` => 500
- `Ticket::Frontend::AgentTicketPhone###RichTextWidth` => 250

chzauleck:
The behaviour of Ticket::Frontend::AgentTicketPhone###RichTextHeight has been fixed for OTOBO 11.1, uncertain if this fix can also be applied to OTOBO 11.0

Ticket::Frontend::AgentTicketPhone###RichTextWidth seems to be working as intended, but has been removed for OTOBO 11.1

stefanhaerter:
I share the observation that support for `RichTextWidth` has seemingly been removed in OTOBO 11.1

Based on that, I furthermore suggest to remove the respective system configuration settings to avoid confusion.

stefanhaerter:
Problem does not persist on 11.0. Closing the issue.

PR #5850, opened 2026-07-13
Issue #3861: Removed remnants of RichTextWidth system configurations.

stefanhaerter:
references #3861
references #5464


stefanhaerter:
Confirmation / review from @chzauleck pending.

PR #5935, opened 2026-07-27
Fixed concatenation issues for translatable strings

urbalazs:
`Translatable()` no-op function does not work with concatenated strings. The `otobo.pot` contains only the first part of the string.

```
#. Perl Module: Kernel/Modules/AdminProcessManagementActivityDialog.pm
msgid "ActivityDialogs currently used in gobal "
msgstr ""

#. Perl Module: Kernel/Modules/AdminProcessManagementActivityDialog.pm
msgid "ActivityDialogs currently used in non-gobal Activities "
msgstr ""
```

This PR fixes the issue.

### DIFF
--- a/Kernel/Modules/AdminProcessManagementActivityDialog.pm
+++ b/Kernel/Modules/AdminProcessManagementActivityDialog.pm
@@ -535,8 +535,7 @@ sub Run {
 
                     $Error{GlobalServerError}        = 'ServerError';
                     $Error{GlobalServerErrorMessage} = Translatable(
-                        'ActivityDialogs currently used in gobal '
-                            . 'Activities may not be set to non-global!'
+                        'Activity dialogs currently used in global activities may not be set to non-global.'
                     );
                 }
                 else {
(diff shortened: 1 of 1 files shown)


stefanhaerter:
I think the core problem is that the regular expressions in the console command do not yet respect that case:

https://github.com/RotherOSS/otobo/blob/ae36955085043dd4d0888dc9eb86bc5fc42299b8/Kernel/System/Console/Command/Dev/Tools/TranslationsUpdate.pm#L276-L281

It should be checked if those regular expressions could be enhanced, in my opinion.

urbalazs:
IMHO removing the concatenation (which is not needed here at all) is a good enough solution. If you change the regex, you may have to handle the unnecessary new line to not put a string to the message catalog something like this:

```
#. Perl Module: Kernel/Modules/AdminProcessManagementActivityDialog.pm
msgid ""
"ActivityDialogs currently used in global \n"
"activities may not be set to non-global."
msgstr ""
```

PR #5963, opened 2026-08-04
Issue #5945: Rebuild customer address via Format function.

### LINKED ISSUE #5945: Ticket split action runs into incorrectly quoted address line phrases.
The parse Module to rebuild address lines has changed between `rel-11_0` and `rel-11_1` from `Mail::Address` to `Email::Address::XS`, which no longer puts smart quotes around phrases that need them.

In the case of ticket splitting, this leads to incorrectly written address lines, that originally pass the `$CheckItemObject`'s checks and only get rebuilt wrongly after it. See below:

Image

stefane2741:
closes #5945

### DIFF
--- a/Kernel/Modules/AgentTicketPhone.pm
+++ b/Kernel/Modules/AgentTicketPhone.pm
@@ -546,8 +546,12 @@ sub Run {
                 $CustomerError = 'ServerError';
             }
 
+            my $Phrase       = $EmailAddressObject->GetRealName( AddressObject => $Email ) || '';
+            my $CustomerKey  = '';
+            my $EmailAddress = $EmailAddressObject->GetAddress( AddressObject => $Email );
+
             # check for duplicated entries
-            if ( defined $AddressesList{$Email} && $CustomerError eq '' ) {
+            if ( defined $AddressesList{$EmailAddress} && $CustomerError eq '' ) {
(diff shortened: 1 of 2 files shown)


Kernel/Modules/AgentTicketPhone.pm
```diff
@@ -581,10 +581,7 @@ sub Run {
                 }
             }
 
-            my $CustomerElement = $EmailAddress;
```
bschmalhofer, line 584: I'm also wondering about line 550, where `$Email` is used as a hash key in the duplicate check.

stefane2741: This check seems a bit weird to me. in line 550, `$Email` is indeed used as the key, but it's not a string, it's a `C<Email::Address::XS>` object.

At the end of the loop at line 601, the correct string variable `$EmailAddress` is used as the key instead. I see no reason why the duplicate check _has_ to come before the extraction of the string email address and why you wouldn't just both times use that actual address.

bschmalhofer: Yes, I think that was simply an oversight when I switched to Email::Address::XS.

bschmalhofer (approved):
The changes look fine to me. Thanks for fixing the messed up lookup in '%AddressesList` .

PR #5967, opened 2026-08-05
Issue #5572: Implemented more stringent checks on content param.

### LINKED ISSUE #5572: VirtualFS Write: Check for `Content` insufficient
In `Kernel::System::VirtualFS::Write()`, the Content is checked like so:

```perl
    # check needed stuff
    for (qw(Filename Content Mode)) {
        if ( !$Param{$_} ) {
            $Kernel::OM->Get('Kernel::System::Log')->Log(
                Priority => 'error',
                Message  => "Need $_!"
            );
            return;
        }
    }
```

However, the POD states:

```perl
=head2 Write()

write a file to virtual file system

    my $Success = $VirtualFSObject->Write(
        Content  => \$Content,
        Filename => '/Object/SomeFileName.txt',
        Mode     => 'binary'            # allowed values are 'utf8' and 'binary'

        # optional, preferences data
        Preferences => {
            ContentType        => 'text/plain',
            ContentID          => '<some_id@example.com>',
            ContentAlternative => 1,
            SomeCustomParams   => 'with our own value',
        },
    );

=cut
```

Observe that `Content` needs to be a reference to a scalar variable. Thus, e.g. an `undef` value would not be `Content => undef`, but rather `Content => \undef` and so bypass the check.

stefanhaerter:
closes #5572

### DIFF
--- a/Kernel/System/VirtualFS.pm
+++ b/Kernel/System/VirtualFS.pm
@@ -199,6 +199,28 @@ sub Write {
         }
     }
 
+    # check content as it should be a reference to a scalar
+    if ( ref $Param{Content} eq 'SCALAR' ) {
+
+        # ensure that it is not undef
+        if ( !defined $Param{Content}->$* ) {
+            $Kernel::OM->Get('Kernel::System::Log')->Log(
+                Priority => 'error',
+                Message  => 'Content is a reference to undef!'
+            );
+
+            return;
+        }
+    }
+    else {
+        $Kernel::OM->Get('Kernel::System::Log')->Log(
+            Priority => 'error',
+            Message  => 'Need Content as reference to scalar!'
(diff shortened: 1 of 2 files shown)


Kernel/System/VirtualFS.pm
```diff
@@ -199,6 +199,28 @@ sub Write {
         }
     }
 
+    # check content as it should be a reference to a scalar
+    if ( ref $Param{Content} && ref $Param{Content} eq 'SCALAR' ) {
```
bschmalhofer: The first check is not needed. `ref` returns an empty string when the argument is not a reference

stefanhaerter: Updated - thank you!

PR #5982, opened 2026-08-06
Issue #5945: Adjust selenium test for new phrase quoting.

### LINKED ISSUE #5945: Ticket split action runs into incorrectly quoted address line phrases.

stefane2741:
closes #5945

### DIFF
--- a/scripts/test/Selenium/Agent/AgentTicketSplit.t
+++ b/scripts/test/Selenium/Agent/AgentTicketSplit.t
@@ -110,7 +110,7 @@ $Selenium->RunTest(
             },
             {
                 SenderType => 'customer',
-                From       => "From Customer <$FromCustomer>",
+                From       => qq{"From Customer" <$FromCustomer>},
                 To         => "To Customer <$ToCustomer>",
             },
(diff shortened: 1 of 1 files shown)


scripts/test/Selenium/Agent/AgentTicketSplit.t
```diff
@@ -110,7 +110,7 @@ $Selenium->RunTest(
             },
             {
                 SenderType => 'customer',
-                From       => "From Customer <$FromCustomer>",
+                From       => "\"From Customer\" <$FromCustomer>",
```
bschmalhofer: The escaped double quotes are not easy on the eye. Could you switch to using `qq{}` ?

Issue #5956, opened 2026-08-04
Remove entrypoint.sh dependency for native systemd file


Splayfery:
At the moment, native installations rely on custom systemd files that are defined in `scripts/systemd`. These are then using the `bin/docker/entrypoint.sh` script to start the Gazelle Perl Webserver on Port 5000.

This dependency is not ideal because it makes the native installations rely on docker specific files and everytime the docker configuration changes the native installations might need to be updated too.

bschmalhofer:
_bin/docker/entrypoint.sh_ is no longer used in the systemd setup. Closing this issue.

PR #5990, opened 2026-08-07
Issue #5956: Bind gazelle webserver port on localhost

Splayfery:
Referring to #5956.

This binds Gazelle webserver Port 5000 from `otobo-web.service` to localhost.
By default, it ensures that the port will never be exposed to other hosts - even when the hosts firewall is absent or misconfigured.

Otobo webservices should only be exposed using a reverse proxy. The user can change this behavior by creating a systemd override file for `otobo-web`. 

This change has been tested on a native Otobo 11.1 installation.


bschmalhofer (approved):
Looks good to me

Splayfery:
I do not have the permission to perform this merge. @bschmalhofer can you do that, please?

Issue #6051, opened 2026-08-24
Use Kernel::System::EmailAddress::Format() in Kernel::System:::CustomerUser backends


bschmalhofer:
A naive regex is used for formatting `UserMailString` in https://github.com/RotherOSS/otobo/blob/1d5ef7f0e1088fbb31d480d68a8e641b006899c4/Kernel/System/CustomerUser/LDAP.pm#L1497 . This looks unsafe when a part of the phrase contains  double quotes.

TODO:

- [x] add a test case where the last name has double quotes
- [x] use Kernel::System::EmailAddress::Format() for creating the full address

bschmalhofer:
Done, closing this issue.

bschmalhofer:
Reopening as Kernel::System:::CustomerUser::DB should also use the Format() method.

bschmalhofer:
Done, the test suite still looks fairly sane, closing this issue again.

PR #6064, opened 2026-08-26
Issue #6051: use the method Format() for UserMailString

bschmalhofer:
in the DB backup. Just like it was already done in the LDAP backend.

### DIFF
--- a/Kernel/System/CustomerUser/DB.pm
+++ b/Kernel/System/CustomerUser/DB.pm
@@ -16,8 +16,11 @@
 
 package Kernel::System::CustomerUser::DB;
 
+use v5.26;
 use strict;
 use warnings;
+use namespace::autoclean;
+use utf8;
 
 # core modules
 use Digest::SHA ();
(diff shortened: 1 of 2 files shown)


Kernel/System/CustomerUser/DB.pm
```diff
@@ -1268,32 +1272,46 @@ sub CustomerUserDataGet {
...
-    my $UserMailString = '';
-    my @UserMailStringParts;
+        my @Values;
 
-    FIELD:
-    for my $Field (@CustomerUserListFields) {
-        next FIELD if !$Data{$Field};
+        FIELD:
+        for my $Field (@CustomerUserListFields) {
+            next FIELD unless $Data{$Field};
```
guest20, line 1293: is 0 a valid value for a field?

svenoe: Changing any behavior around which fields are considered is not part of this PR. (And disregarding `0` as first name, last name or email address (in most cases) will make sense probably almost all of the times, I guess.)

guest20: Fair

PR #6066, opened 2026-08-27
Issue #5897: Extend html support to p, img and table cells.

### LINKED ISSUE #5897: CKEditor 5: GeneralHtmlSupport allow list is hardcoded, stripping e.g. image border styles
### Environment
- OTOBO 11.0.15 / current `rel-11_1`
- CKEditor 5 (47.6.1 bundled)

### Problem
The `htmlSupport` (GeneralHtmlSupport) allow list of the RichText editor is hardcoded in `Core.UI.RichTextEditor.js`:

```js
htmlSupport: {
    allow: [
        { name: 'span',  attributes: true, classes: true, styles: true },
        { name: 'cite',  attributes: true, classes: true, styles: true },
        { name: 'style', attributes: true, classes: true, styles: true },
        { name: 'table', attributes: true, classes: true, styles: true },
    ],
    ...
}
```

Any attribute/class/inline style on other elements that is not covered by an active editor feature is silently stripped, and there is no configuration to change that.

### Reproduction example
1. Open any RichText editor in the agent interface (e.g. AgentTicketNote).
2. Switch to source editing and add ``, or paste HTML containing such an image.
3. The `border` style is removed from the content (`editor.getData()` contains no `border`).

This is a regression compared to OTOBO 10 / CKEditor 4, where agents could give images a border via the image dialog, and such content survived round-trips. It also affects `td`/`p`/`div` level styling pasted from external sources.

### Proposal
Move the element list into SysConfig, defaulting to the current behavior:

- `Frontend::RichText::HtmlSupportAllowedElements` (agent), default: `span, cite, style, table`
- `CustomerFrontend::RichText::HtmlSupportAllowedElements` (customer), same default

`Layout.pm` passes the list through the existing `RichText` JS data, `Core.UI.RichTextEditor.js` builds the allow array from it. Installations keep today's filtering unless an admin extends the list (e.g. adds `img`).

A pull request implementing this is ready and will be linked here.

stefane2741:
closes #5897

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.RichTextEditor.js
+++ b/var/httpd/htdocs/js/Core.UI.RichTextEditor.js
@@ -192,6 +192,18 @@ Core.UI.RichTextEditor = (function (TargetNS) {
             htmlSupport: {
                 allow: [
                     {
+                        name: 'style',
+                        attributes: true,
+                        classes: true,
+                        styles: true
+                    },
+                    {
+                        name: 'p',
+                        attributes: true,
+                        classes: true,
+                        styles: true
+                    },
+                    {
                         name: 'span',
                         attributes: true,
                         classes: true,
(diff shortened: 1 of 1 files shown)


stefane2741:
Before merging, it would probably be advisable to think about whether any of these rather permissible extensions of the supported HTML pose a potential XSS vulnerability?

chzauleck (approved):

Issue #5972, opened 2026-08-05
Daemon-forked Console::Command tasks (e.g. Maint::Ticket::EscalationCheck) crash in DBD::MariaDB disconnect_all() during global destruction (OTOBO 11.1.0 Beta1, DBD::MariaDB 1.24)


hndrx-commits:
## Summary

After upgrading a test instance from OTOBO 11.0.5 to 11.1.0 Beta1, every scheduled
task executed by the internal Daemon (`SchedulerCronTaskManager` / `SchedulerTaskWorker`)
reliably triggers a Perl segfault at process exit, inside `DBD::MariaDB`'s
`disconnect_all()` implementation. The scheduled task itself completes successfully;
the crash happens afterwards, during Perl global destruction.

This was never observed before the 11.1 upgrade on the same host/OS/library stack.

## Environment

- OTOBO version: 11.1.0 Beta1 (upgraded from 11.0.5 via `DBUpdate-to-11.1.pl`)
- OS: CentOS Stream 9
- Deployment: Apache/mod_perl (no otobo-web systemd unit), daemon via `otobo.Daemon.pl`
- DB backend: MariaDB server 10.5.29-4.el9
- `mariadb-connector-c`: 3.2.6-1.el9 (no newer version available in AppStream for this OS release)
- `DBD::MariaDB`: **1.24**, installed under a local::lib path
  (`/opt/otobo/install/local/lib/perl5`), confirmed via gdb backtrace as the version
  actually loaded by the Daemon process (system RPM `perl-DBD-MariaDB-1.21-17.el9`
  is also present but not the one loaded here)
- Perl: 5.32.1

## Steps to reproduce

1. Fresh install or upgrade to OTOBO 11.1.0 Beta1 on CentOS Stream 9 with the
   above package versions.
2. Leave the default Daemon Scheduler Cron Task `EscalationCheck`
   (`Daemon::SchedulerCronTaskManager::Task###EscalationCheck`, schedule `*/5 * * * *`,
   module `Kernel::System::Console::Command::Maint::Ticket::EscalationCheck`) active
   (default/shipped task, not custom).
3. Let `otobo.Daemon.pl` run normally.
4. Observe `journalctl -k` / `coredumpctl list` every 5 minutes.

## Observed behavior

Every ~5 minutes, in exact time correlation with the `EscalationCheck` task log entry
in `SchedulerTaskWorker*.log`, the kernel logs a segfault of the forked worker process:
[...]

Issue #1471, opened 2021-12-01
Elasticsearch self repair and handling of "valid" error responses


svenoe:
If for some reason the creation of a ticket in Elasticsearch fails, all subsequent article creations and other ticket updates will fail. We receive a clear error response though, and should use it to try to create the ticket a second time. While at it, we could filter out errors from not being able to handle specific files (like encrypted files, etc.) from the ingest plugin, and maybe put an info text to the log, but not more.

For both to work, first the invokers must be extended to handle errors (task five of https://github.com/RotherOSS/otobo/issues/772). The important places for this would be:
 - Kernel/GenericInterface/Requester.pm line 324: `$FunctionResult = $TransportObject->RequesterPerformRequest(`
 - Kernel/GenericInterface/Transport/HTTP/REST.pm line 872: `$RestClient->$RestCommand(@RequestParam);`
I'm not sure how to do this in a nice way, without looking further into this. In principle we don't want the error handling right after the second, but the communication with the invoker usually is done in the first (e.g. line 206: `my $FunctionResult = $InvokerObject->PrepareRequest(` and especially interesting since we would probably need something similar line 226: `elsif ( $FunctionResult->{StopCommunication}`).

For the self repair of the ticket index, line 539-552 of Kernel/System/Console/Command/Maint/Elasticsearch/Migration.pm `# create the ticket` could be taken, which is quite simple in itself, some measurement has to be taken, though, to not cause infinite loops! :)

bschmalhofer:
@svenoe 
I started looking into this. My understanding is that there already is generic support for error handling. That is line 619 in  _Kernel/GenericInterface/Requester.pm:

>     my $InvokerHandleErrorOutput = $Param{InvokerObject}->HandleError(
>         Data => $HandleErrorData,
>     );

So the first line of investigation would be adding a `HandleError()` method in _Kernel/GenericInterface/Invoker/Elasticsearch/TicketManagement.pm_ .
I was wondering how to provoke an error in submitting the ticket to ES. The most simple idea I had, is to shut down ES temporarily. Is there a better approach?

svenoe:
[...]

PR #6101, opened 2026-09-09
Fix Elasticsearch missing-document recovery for typeless errors

frankii91:
Updating a ticket whose Elasticsearch document is missing should recreate the ticket and its articles. With Elasticsearch 9.4.4, the existing recovery path passes an undefined TicketID because it only recognizes the older `[_doc][5]: document missing` reason, while Elasticsearch now returns `[5]: document missing`.

Accept both reason formats and stop recovery with a clear error when a valid TicketID cannot be extracted, before calling ticket or article APIs. The existing recovery sequence is otherwise unchanged. Add an isolated regression test covering both formats, malformed/missing IDs, ticket and article recovery calls, and ordinary/transport responses.

Validation:
- The regression test fails on the original module for the newer response format and passes with this patch (12 top-level TAP tests, including subtests; no database or Elasticsearch access).
- Tested on OTOBO 11.1.0-beta2 with Elasticsearch 9.4.4 using the same module as current `rel-11_1`. Before the patch, a synchronous TicketManagement update for an actually missing document returned Success=1, logged three `Need TicketID!` errors, and left the document missing.
- After patching the module, the same update triggered the existing self-repair path successfully. All 18 affected tickets were recovered through OTOBO, with all 37 article entries verified against database counts. Repeating an update on one recovered ticket did not duplicate its articles. No direct document insertion or full index rebuild was used.
- The full application test suite was not run on the live installation.

Related to the self-repair behavior introduced for #1471. We discovered these index gaps while investigating an installation affected by the separate crash issue #5972. The missing tickets dated from the earlier failure period, but this does not establish that the driver crash directly corrupted Elasticsearch: the defect fixed here is the failure to recover a missing document, regardless of how it became missing.

### DIFF
--- a/Kernel/GenericInterface/Invoker/Elasticsearch/TicketManagement.pm
+++ b/Kernel/GenericInterface/Invoker/Elasticsearch/TicketManagement.pm
@@ -881,11 +881,20 @@ sub HandleResponse {
     #     "type":"document_missing_exception",\
     #     "reason":"[_doc][5]: document missing","index_uuid":"rH43T-SsTz-H_k8aFg13dQ","shard":"0","index":"ticket"},"status":404}
     if ( $Param{Data}->{error} && $Param{Data}->{error}->{type} eq 'document_missing_exception' ) {
+        # Elasticsearch can report either "[_doc][5]: document missing" or "[5]: document missing".
+        my $Reason = $Param{Data}->{error}->{reason} // '';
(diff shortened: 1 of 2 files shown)


bschmalhofer:
This looks good. I like the creative approach taken in the test scripts.

bschmalhofer:
The PR is merged. After the merge I ran CodePolicy over the changed and created files.

PR #6132, opened 2026-09-15
Issue #6131: Pass on ACLHidden if not visibile.

### LINKED ISSUE #6131: DynamicField Set: Mandatory inner fields in hidden Sets prevent form submit in AgentTicketProcess
Reason is that `ACLHidden` is only passed on if the set itself is both invisible and mandatory. This should be aligned with the behavior in `EditSectionRender`, and CustomerTicketProcess should be checked as well.

stefanhaerter:
closes #6131


MichaelThumes (approved):
lgtm

PR #6133, opened 2026-09-15
Issue #5510: Fixed popup handling for AgentTicketMove.

### LINKED ISSUE #5510: can't "cancel and close" AgentTicketMove
When trying to "Cancel & Close" AgentTicketMove after opening it on a ticket that has already been locked, nothing happens.

I suspect this to be due to the fact that AgentTicketMove does not open in a popup, but uses the same template structure, causing the "cancel & close" button to behave differently.

I have not checked (yet) whether this issue is already present in OTOBO 10.1, but it occurs in 11.0 and 11.1

stefanhaerter:
closes #5510


chzauleck (approved):

MichaelThumes (approved):
looks good to me

Issue #6010, opened 2026-08-12
Elasticsearch Abbrevation on the Resultpage too aggressive


73:
### Environment

- OTOBO version: 11.1.x

### Expected behavior

I expect that table columwidth is determined by text shown.
As much text is shown as possible.

### Actual behavior

The error occurs on e.g. long titles of FAQ on the ES resultpage (after pressing enter on the quick result modal).

Table entries are abbrevated while there is a lot of space in the cell.

### How to reproduce

1. Create two FAQ with a long and medium title
2. search by ES
3. press enter

PR #6139, opened 2026-09-18
Issue #6010: Removed truncation in tt

rmfalves:


MichaelThumes (approved):
tested with a over-long ticket title on FF and chromium, and it overflows/truncates nicely.

less is more.

PR #6140, opened 2026-09-18
Set configurable search limit in Customer interface

rmfalves:


Kernel/Config/Files/XML/Framework.xml
```diff
@@ -8943,6 +8943,13 @@ via the Preferences button after logging in.
         <Value>
             <Item ValueType="Checkbox">1</Item>
         </Value>
+    </Setting>
+    <Setting Name="AdminCustomerUser::SearchLimit" Required="1" Valid="1">
```
stefanhaerter: Is there a specific reason for the setting being required, if a hardcoded fallback does exist?

Kernel/Config/Files/XML/Framework.xml
```diff
@@ -8943,6 +8943,13 @@ via the Preferences button after logging in.
...
+        <Description Translatable="1">Controls the search limit in the AdminCustomerUser interface.</Description>
+        <Navigation>Frontend::Admin::View::CustomerUser</Navigation>
+        <Value>
+            <Item ValueType="String" ValueRegex="^[0-9]{1,6}$">200</Item>
```
stefanhaerter, line 8951: Why do you require 1 to 6 digits? Doesn't that unnecessarily prevent leaving it empty?

stefanhaerter:
Besides my above remarks, I'd either implement this for rel-10_1 or rel-11_1 - but that's only my opinion.

Kernel/Config/Files/XML/Framework.xml
stefanhaerter: I withdraw that comment, as it seems to be a common pattern in OTOBO system configuration settings.

Kernel/Config/Files/XML/Framework.xml
```diff
@@ -8943,6 +8943,13 @@ via the Preferences button after logging in.
         <Value>
             <Item ValueType="Checkbox">1</Item>
         </Value>
+    </Setting>
+    <Setting Name="AdminCustomerUser::SearchLimit" Required="0" Valid="1">
+        <Description Translatable="1">Controls the search limit in the AdminCustomerUser interface.</Description>
```
gyarakilaszlo, line 8948: Please use the real name of the view instead of the technical file name.

stefanhaerter: The name is similar to other settings targeting the same mask. I'd rather not break the pattern for a single setting.

MichaelThumes (approved):
looks good to me as well

Issue #5124, opened 2026-02-05
Wrong hint on Queue Management screen for setup screens


urbalazs:
### Expected behavior

Admins should find the relevant configuration screen by searching for its name.

### Actual behavior

Queue Management page shows wrong module names as setting explanation:

- System Address: Is defined in Admin > System addresses, but system address can be modified in Admin > Email Addresses.
- Calendar: Is defined in Admin > SystemConfiguration > Core > Time, but the correct wording is System Configuration (with a space between the words).

Based on the explanations, admins cannot find the relevant configuration screens by using the search field and entering (or copy-pasting) the name as is in the settings explanation. 

### How to reproduce

1. Go to Queue Management screen and add a new queue.
2. Scroll down to setting "System Address" or "Calendar".
3. Remember the name (or copy it), go to admin screen and enter the name in the search field. There is no result due to the wrong name in the explanation.

stefanhaerter:
Putting this into the OTOBO Issues project for triage.

svenoe:
(The first point is already changed for OTOBO 11.1 at least.)

PR #6141, opened 2026-09-18
Issue #5124: Fixed wrong hint in queue management

rmfalves:


stefanhaerter (commented):
Looks good. Question would be if we want to change this for rel-10_1 already.

MichaelThumes (approved):

